PR #1, opened 2019-11-27
Move over all work in progress for doc conversion

agjohnson:
This PR handles almost all of the doc conversion into reST, PDF and CSS fixes, and some Sphinx customization changes. There are a few pieces missing currently:

* There were a couple of places where a conditional `condition=isSandbox` or `condition=isStandalone` was used. However, separation of the Application Dev Guide and the other guides breaks usage of conditional content in the application dev guide. An example of this is here: https://github.com/MicroEJ/docs/pull/1/commits/236b9b906ba166f2a42420abebcd10b7250f3d27#diff-afcf02fa31b18d85009534434baa7252R58
* I have brought up a small change to the structure of the TOC tree in https://github.com/agjohnson/microej-docs/pull/1
* There is a separate branch that includes all of the fixes above and was forked right before removing the application dev guide from standalone/sandbox guides. This branch will exist as an example of using multiple projects for each of these subpaths


agjohnson:
I need to merge this into master to complete set up of RTD versions. I will squash merge this for now. I've pushed up commits a18606c...4009bb4 in the branch `pr/1`.

PR #28, opened 2020-04-13
Use "abstraction layer" instead of "adaptation layer"

jerome-leroux:


frivieremicroej:
OK - platform schematics were already updated

PR #78, opened 2020-09-29
Update schemas of MEJ Firmware tutorial

gjadi-mej:


thomasponsmicroej:
Ok with the schema content, about the design two small issues I guess:
- it seems that in the first diagram the title of the right panel is on two lines instead of one line: "MicroEJ Resources". Is it normal? It feels strange because there is a blank space + small padding with the text below. 
- Same with the second diagram, there is a return to line on one of the arrow "(.java, .c)" between "Foundation Libraries Mockups" and "Java/C Compiler"

thomasponsmicroej (approved):
OK

PR #129, opened 2020-12-02
Document visiblity of mock methods

gjadi-mej:


gjadi-mej:
https://docs.microej.com/en/gjadi-mock-unsatisfiedlinkerror/PlatformDeveloperGuide/mock.html#example

At the very end of the example section.

PR #138, opened 2020-12-09
M0228J-74 Add getting started and remove empty chapters

tdelhomenie-microej:
This change adds a getting started and remove empty chapters.

I am not sure the use of the Standalone Application in the getting started is the right choice:
- is it the most commonly used application type for starting ? What about a Java lib ?
- it requires to add the platform before the build since it builds the firmware
- it fails if the property "sp.server.port" is not set (not in the default common.properties)

Readthedocs link: https://docs.microej.com/en/feature-m0228j-74_clean-doc/


ApplicationDeveloperGuide/microjs/getting_started.rst
```diff
@@ -9,20 +9,67 @@
...
+.. code:: java
+
+    import com.microej.microjs.JsErrorWrapper;
+    import com.microej.microjs.MicroJsCode;
+    import com.microej.microjs.MicroJsRuntime;
+
+    ...
+
+    MicroJsCode.initJs();
+    MicroJsRuntime.ENGINE.run();
```
bguedas, line 34: The `stop()` below won't be called because `run()` never returns. You might use `MicroJsRuntime.ENGINE.runOneJob()`.

ApplicationDeveloperGuide/microjs/getting_started.rst
```diff
@@ -9,20 +9,67 @@
...
 ===============
 
-Installation
-------------
+Let's walk through the steps required to use Javascript in your MicroEJ application:
 
-Project structure
------------------
+- create your application project in the MicroEJ SDK, see :ref:`standalone_application`. 
+  Once done, you can continue with the MicroEJ SDK or with your favorite IDE.
```
bguedas, line 15: not a list item (missing `-`)?

ApplicationDeveloperGuide/microjs/getting_started.rst
```diff
@@ -9,20 +9,67 @@
...
+    import com.microej.microjs.MicroJsCode;
+    import com.microej.microjs.MicroJsRuntime;
+
+    ...
+
+    MicroJsCode.initJs();
+    MicroJsRuntime.ENGINE.run();
+    MicroJsRuntime.stop();
+
+- create a file named ``hello.js`` in the folder ``src/main/js`` with the following content:
```
bguedas, line 37: `src/main/js` should be the right folder, but at the moment it is still `src/main/resources`.

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -255,6 +255,175 @@ and this :ref:`tutorial <tutorial_setup_automated_build_using_jenkins_and_artifa
...
+
+The MicroEJ SDK also provides a new Command Line Interface (CLI).
+It allows to perform all the main development operations without the MicroEJ SDK.
+This CLI is the good tool if you want to work in a terminal or in any other IDE.
+
+The following operations are supported:
+
+- building a module project
+- running the project application on the simulator
+- publishing a module in a module repository
```
bguedas, line 271: Shouldn't we end each list item with `;` and the last one with `.`?

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -255,6 +255,175 @@ and this :ref:`tutorial <tutorial_setup_automated_build_using_jenkins_and_artifa
...
+
+   Target "simulator:run" does not exist in the project "my-app".
+
+it means that the subcommand ``run`` is not supported by the build type of your application.
+Make sure it is one of the following ones:
+
+- ``build-application``, with version ``7.1.0`` or higher
+- ``build-microej-javalib``, with version ``4.2.0`` or higher
+- ``build-firmware-singleapp``, with version ``1.3.0`` or higher
+
```
bguedas, line 426: I guess that all the text above is a copy of the REAME. Should the doc refer to the README or the README to the doc?

The benefit of the README is that it is versioned with the code.

bguedas (changes requested):
I've not tested the HTML page layout yet.

ApplicationDeveloperGuide/microjs/getting_started.rst
tdelhomenie-microej: What is the best ? run wihtout stop or runOneJob with stop ?

ApplicationDeveloperGuide/microjs/getting_started.rst
tdelhomenie-microej: No, I go to a new line to avoid having a very long line, but in HTML it is on the same line.

bguedas: Ok, I was not sure

ApplicationDeveloperGuide/mmm.rst
tdelhomenie-microej: The practice is to either use `.` or nothing, but most importantly to be consistent. Is there a convention at MicroEJ about this ?

ApplicationDeveloperGuide/mmm.rst
tdelhomenie-microej: Right, mostly copy/paste. I agree keeping the README sounds better, but I was not sure it is a common practice to refer to a README in the doc. Let's go to make a link to the README ?

ApplicationDeveloperGuide/microjs/getting_started.rst
bguedas: I think that `MicroJsRuntime.ENGINE.runOneJob()` with `MicroJsRuntime.stop()` is the best option for a HelloWorld: the former executes the only scheduled job whereas the latter stops the timer (otherwise there is a time scheduler thread that continues).

tdelhomenie-microej: ok. Done.

ApplicationDeveloperGuide/microjs/getting_started.rst
tdelhomenie-microej: Thanks. Done.

ApplicationDeveloperGuide/mmm.rst
bguedas: Okay, you're right. This is the practice in english.

ApplicationDeveloperGuide/mmm.rst
bguedas: I don't know, it was just to start the discussion :) We may ask the others their opinion

ApplicationDeveloperGuide/microjs/getting_started.rst
dalgassimou: How about changing in the runOneJob ==> to run and run ==> runUntilStopped
The run method will run all jobs and stop itself when there is no job in it queue (it'll also call the engine.stop to clean up)
and runUntilStopped will run until someone call the stop method.

ApplicationDeveloperGuide/mmm.rst
bguedas: as discussed in meeting, the full explainations are kept in the doc.

tdelhomenie-microej: I will push a MR with the update of the README.

tdelhomenie-microej: https://gitlab.cross/M0090_IDE/M0090_MMM-CLI/-/merge_requests/10

bguedas (approved):

PR #139, opened 2020-12-10
Feature/m0078 ddg 355 improve mmm doc

gjadi-mej:
- how to enable MMM semantic to a module
- how to set MMM matchinig rule on a dependency
- how to enable icy-update easyant plugin
- how to set dependency visibility on a dependency

### DIFF
--- a/ApplicationDeveloperGuide/mmm.rst
+++ b/ApplicationDeveloperGuide/mmm.rst
@@ -154,7 +154,7 @@ Set the matching rule of a given dependency with ``ej:match="matching rule"``.  
 
 .. code-block:: xml
 
-   <depedency org="ORG" name="NAME" rev="MAJOR.MINOR.PATCH" ej:match="perfect" />
+   <dependency org="ORG" name="NAME" rev="MAJOR.MINOR.PATCH" ej:match="perfect" />
 
 Automatic Update Before Resolution
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -172,6 +172,33 @@ When the plugin is enabled, for each module dependency, MMM will check
 the version declared in the module file and update it to the latest
 version available if and only if it satisfies the matching rule of the
 dependency.
+
+Dependency Visibility
+~~~~~~~~~~~~~~~~~~~~~
+
+- A dependency declared ``public`` is transitively resolved by upper
+  modules.  The default when not set.
+- A dependency declared ``private`` is only used by the module itself,
+  typically for:
+
+   - Bundling the content into the module
+   - Testing the module
+
+The visibility of a dependency is declared by setting the tag
+``visibility``, for example:
+
+.. code-block:: xml
+
+   <dependency org="ORG" name="NAME" rev="MAJOR.MINOR.PATCH" visibility="private" />
+
+The visibility can also be set according to the configurations
+declared in the ``configurations`` node.  The configuration is
+declared by setting the ``conf`` tag, for example:
+
+.. code-block:: xml
+
+   <dependency org="ORG" name="NAME" rev="MAJOR.MINOR.PATCH" conf="[CONF-NAME]->*" />
+
 
 .. _mmm_configuration:
 
--- a/ApplicationDeveloperGuide/mmm.rst
+++ b/ApplicationDeveloperGuide/mmm.rst
@@ -48,7 +48,7 @@ Specification
 
 MMM provides a non ambiguous semantic for dependencies
 resolution. Please consult the MMM specification available on
-`<https://developer.microej.com/packages/documentation/TLT-0831-SPE-MicroEJModuleManager-2.0-D.pdf>`_.
+`<https://developer.microej.com/packages/documentation/TLT-0831-SPE-MicroEJModuleManager-2.0-E.pdf>`_.
 
 
 .. _mmm_module_skeleton:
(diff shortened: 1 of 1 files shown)


tdelhomenie-microej:
Most of the content added in the PR is already in the spec https://repository.microej.com/packages/documentation/TLT-0831-SPE-MicroEJModuleManager-2.0-D.pdf which is referenced in the documentation here : https://docs.microej.com/en/latest/ApplicationDeveloperGuide/mmm.html#specification
It would duplicate the information.
Actually I think we should move all the relevant content of the pptx in the doc directly and remove the pptx. It would make the content searchable.
WDYT ?

gjadi-mej:
I disagree.

If some information about how to **use** MMM is missing from the doc, then it should be added.  But the specification document may (and probably should) contain more details.  IMO the MMM Spec is similar to the B-ON or SNI specification.

For example, I don't think it is necessary to add to docs.microej.com all the examples about which dependencies are fetched depending of the matching rules.  Yet it is valuable to have it.  (Though, it could be added in an Appendix).

tdelhomenie-microej:
The "spec" mixes information types, it contains also usage documentation, that's why most of the content of the PR is already in the MMM spec. My initial concern is that it duplicates the information and it is always difficult to maintain consistency, that's why I proposed a way to avoid that.
Having resolution examples in the doc sounds good to me, it helps to understand how it works and to troubleshoot. Maven or gradle for example have detailed documentation about dependency resolution. The doc could be split in several parts to make it easy to read (usage / reference ?).

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
        </dependencies>
    </ivy-module>        
 
+Enable MMM Semantic
+~~~~~~~~~~~~~~~~~~~
+
+The MMM semantic is enabled in a the module by adding the MicroEJ XML
```
tdelhomenie-microej, line 116:
```suggestion
The MMM semantic is enabled in a module by adding the MicroEJ XML
```

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
...
+namespace in the ``ivy-module`` node:
```
tdelhomenie-microej, line 117:
```suggestion
namespace and the ``ej:version`` attribute in the ``ivy-module`` node:
```

tdelhomenie-microej (changes requested):

ApplicationDeveloperGuide/mmm.rst
gjadi-mej: thanks.

ApplicationDeveloperGuide/mmm.rst
gjadi-mej: Thanks.

gjadi-mej:
I agree with your reasoning but this is a bigger rework.
Until we decide how to present the spec, I think this PR makes it better _now_. :)

tdelhomenie-microej:
ok, if @frivieremicroej agrees, let's merge this PR and open a ticket for the bigger rework. I will try to find the time to do it soon.

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
...
+namespace and the ``ej:version`` attribute in the ``ivy-module`` node:
+
+.. code:: xml
+
+   <ivy-module xmlns:ej="https://developer.microej.com" ej:version="2.0.0">
+
+.. note::
+
+   Multiple namespaces can be declared in the ``ivy-module`` node.
+
```
frivieremicroej, line 126: Indicate this is automatically set by Module Nature Skeletons

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
...
+
+Dependency Matching Rule
```
frivieremicroej, line 128: Module Dependency

- Add the default line and explain this is "compatible"
- Matchind Rule
- Visibility

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
...
+   <info organisation="[organisation]" module="[name]" status="integration" revision="[version]">
+       <ea:plugin org="com.is2t.easyant.plugins" name="ivy-update" revision="1.+" />
+   </info>
+
+When the plugin is enabled, for each module dependency, MMM will check
+the version declared in the module file and update it to the latest
+version available if and only if it satisfies the matching rule of the
+dependency.
+
+Dependency Visibility
```
frivieremicroej, line 176: moved to Module Denpendency

ApplicationDeveloperGuide/mmm.rst
```diff
@@ -110,6 +110,96 @@ It describes the :ref:`module nature <module_natures>` (also called build type)
...
+.. code-block:: xml
+
+   <info organisation="[organisation]" module="[name]" status="integration" revision="[version]">
+       <ea:plugin org="com.is2t.easyant.plugins" name="ivy-update" revision="1.+" />
+   </info>
+
+When the plugin is enabled, for each module dependency, MMM will check
+the version declared in the module file and update it to the latest
+version available if and only if it satisfies the matching rule of the
+dependency.
```
frivieremicroej, line 174:
```suggestion
the version declared in the module file and update it to the highest
version available which satisfies the matching rule of the
dependency.
```

frivieremicroej:
> ok, if @frivieremicroej agrees, let's merge this PR and open a ticket for the bigger rework. I will try to find the time to do it soon.

Goal is move all Ivy specific stuff to MMM docs and then later remove from the PDF specification

PR #143, opened 2020-12-14
Add limitations

plaupretre:


ApplicationDeveloperGuide/microjs/limitations.rst
```diff
@@ -10,6 +10,20 @@
 Limitations
 ===========
 
+Syntaxes
+********
+
+The following syntaxes are not supported by the MicroEJ JavaScript engine:
+
+- ``with (x) { }``
+- ``break`` or ``continue`` in an unreachable section of code i.e. after a ``return`` or another ``break`` or ``continue``
```
bguedas, line 19: it's not only `break` or `continue` in an unreachable section of code (it is the case of every statement in unreachable section). I'll simply say "unreachable code after `return`, `break` , `continue`` (or `throw`)"

ApplicationDeveloperGuide/microjs/limitations.rst
```diff
@@ -10,6 +10,20 @@
 Limitations
 ===========
 
+Syntaxes
```
bguedas, line 13: Is it really correct to name it "syntaxes". What do you think of "Unsupported Statements"?

bguedas (changes requested):

PR #162, opened 2021-01-06
remove the default conf when fetching architecture

frivieremicroej:
- useless with architectures uploaded to repository.microej.com up to
7.14.0 (without ivy.xml) - was not used in Wrover Platform
https://github.com/MicroEJ/Platform-Espressif-ESP-WROVER-KIT-V4.1/blob/master/ESP32-WROVER-Xtensa-FreeRTOS-configuration/module.ivy#L31
- useless and fails download of architecture 7.15.0 with ivy.xml and new
naming convention (no default conf)


gjadi-mej (approved):
Ok.

PR #191, opened 2021-02-11
document run.bat

frivieremicroej:


PlatformDeveloperGuide/platformCreation.rst
```diff
@@ -498,6 +498,33 @@ The build script must implement the following specification:
...
+- On success
+
+  - The executable has been uploaded and started on the device
+  - The script must end with zero exit code.
+
+The run script can optionally redirect execution traces. If it does not implement execution traces redirection,
+the testsuite must be configured with the following :ref:`application_options` in order to take its input from a TCP/IP socket server, 
+such as :ref:`tool_serial_to_socket`.
+
+::
```
gjadi-mej, line 522:
```suggestion
.. code-block:: properties
```

PlatformDeveloperGuide/platformCreation.rst
```diff
@@ -498,6 +498,33 @@ The build script must implement the following specification:
...
+- On Mac OS X or Linux operating systems, it is a shell script named ``run.sh``, with execution permission enabled.
+- On build error, the script must end with a non zero exit code.
+- On success
+
+  - The executable has been uploaded and started on the device
+  - The script must end with zero exit code.
+
+The run script can optionally redirect execution traces. If it does not implement execution traces redirection,
+the testsuite must be configured with the following :ref:`application_options` in order to take its input from a TCP/IP socket server, 
+such as :ref:`tool_serial_to_socket`.
```
gjadi-mej, line 520: Add a link to Testsuite Engine ?

And I don't know how, but we should explain more how to use the trace analyzer without the Serial To Socket.

PlatformDeveloperGuide/platformCreation.rst
```diff
@@ -498,6 +498,33 @@ The build script must implement the following specification:
 Many build script templates are available for most commonly used C toolchains in the 
 `Platform Qualification Tools repository <https://github.com/MicroEJ/PlatformQualificationTools/tree/master/framework/platform/scripts>`_.
 
+Run Script File
+---------------
+
+This script is required only for Platforms intended to run a :ref:`MicroEJ Testsuite <platform_testsuite>` on device.
+
+The BSP run script is responsible to invoke a third-party tool to upload and start the executable file (``application.out``) on device.
```
jcourcelmicroej, line 506: Here we do not indicate from where the executable file ``application.out`` comes from and where it's located exactly when the ``run.bat`` or ``.sh`` script is called. It's important to indicates that this file has to be available in the working directory.

PlatformDeveloperGuide/platformCreation.rst
frivieremicroej: 1) The testsuite engine is wrapped by the MMM testsuite plugin layer then wrapped by the the PQT testsuites layers?

The mode is documented here https://github.com/MicroEJ/PlatformQualificationTools/blob/master/tests/fs/java/java-testsuite-fs/config.properties.tpl#L50. It is enabled by undocumented option 'launch.test.trace.file' https://github.com/MicroEJ/PlatformQualificationTools/blob/master/tests/fs/java/java-testsuite-fs/module.ant#L12

I think this is sufficient so far.

PlatformDeveloperGuide/platformCreation.rst
frivieremicroej: done

PR #192, opened 2021-02-16
Disable transitive resolution of the architecture for PCA platform

gjadi-mej:
This prevents "Too many loaded Platforms" error in `platform-loader`.


frivieremicroej:
See cross comment won't

PR #208, opened 2021-03-04
Explain how to add the ivy `repository` configuration when embedding a module in a repository

gjadi-mej:
The ivy `repository` configuration is required as per the declaration
above, but this configuration is not (yet) created by default in the
skeleton.

### DIFF
--- a/ApplicationDeveloperGuide/repository.rst
+++ b/ApplicationDeveloperGuide/repository.rst
@@ -110,8 +110,20 @@ The ``artifacts`` configuration has to be derived with a new name as many times 
 Include a Module Repository
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To add all the modules already included in an other module repository, 
-declare the module repository dependency using the ``repository`` configuration:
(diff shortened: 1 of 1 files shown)


ApplicationDeveloperGuide/repository.rst
```diff
@@ -122,6 +122,17 @@ declare the module repository dependency using the ``repository`` configuration:
       <!-- ... other dependencies ... -->
    </dependencies>
 
+Add the configuration ``repository`` if it does not exist:
```
frivieremicroej, line 125: I prefer to invert the steps (in the file modification order)

```
To add all the modules already included in an other module repository:

- declare the configuration ``repository`` if it does not exist:

XXX

- declare the repository module dependency:

XXX

```

frivieremicroej (changes requested):

ApplicationDeveloperGuide/repository.rst
gjadi-mej: updated

PR #212, opened 2021-03-15
Update create firmware from scratch tutorial

microej-czaepfel:


Tutorials/tutorialCreateFirmwareFromScratch.rst
```diff
@@ -888,11 +887,11 @@ C header ``sni.h``.
      	/* Configure the clocks, UART and GPIO. */
      	prvSetupHardware();
      
-        
 
 Build and Link the Firmware with the MicroEJ Runtime and MicroEJ Application
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+To build and link the firmware with the MicroEJ Runtime and MicroEJ
```
gjadi-mej, line 894: Duplicate line.

Tutorials/tutorialCreateFirmwareFromScratch.rst
```diff
@@ -552,11 +532,29 @@ invoking ``make`` in the FreeRTOS BSP.
      # Specify MicroEJ Platform header files ('*.h') parent directory.
      # This is a '/' separated directory relative to 'bsp.root.dir'.
      microejinc.relative.dir=microej/inc
+	 
+#. Edit the file ``modules.ivy`` and add the MicroEJ architecture as a dependency:
```
gjadi-mej, line 536:
```suggestion
#. Edit the file ``modules.ivy`` and add the MicroEJ Architecture as a dependency:
```

Tutorials/tutorialCreateFirmwareFromScratch.rst
```diff
@@ -552,11 +532,29 @@ invoking ``make`` in the FreeRTOS BSP.
...
 
-    .. image:: images/tuto_microej_fw_from_scratch_build_platform.PNG
+	 # Platform configuration file (relative to this project).
+	 com.microej.platformbuilder.platform.filename=Tuto.platform
+
+#. Right-click on the platform project and click on ``Build Module``. 
+
+#. Select 'File' > 'Import...' > 'General' > 'Existing Projects into Workspace' > 'Next'.
+
+#. Check 'Select root directory' and import the generated platform in the workspace 'lm3s811evb-Platform-CM0_GCC48-0.0.1'.
```
gjadi-mej, line 557: Use `:guilabel:`.
Can you also add the output generated at the end of the build that mentions the path?
Ensure that the path is consistent with other paths used in the tutorial.

Tutorials/tutorialCreateFirmwareFromScratch.rst
```diff
@@ -266,33 +266,14 @@ the UART device and an example implementation for QEMU is available
...
-    int puts(const char *s) {
-        while (*s) {
-            putchar(*s);
-            s++;
-        }
-        return putchar('\n');
-    }
-
-And here is the patch that implements both functions and prints
+Here is the patch that implements ``putchar(3)`` and ``puts(3)`` and prints
```
gjadi-mej, line 269: Ok with the removal of this duplicate.  Good catch.

Tutorials/tutorialCreateFirmwareFromScratch.rst
gjadi-mej: Also, add a link to https://docs.microej.com/en/latest/PlatformDeveloperGuide/platformCreation.html#microej-platform-build
and a screenshot with the platform imported into the workspace (as before).

Tutorials/tutorialCreateFirmwareFromScratch.rst
```diff
@@ -535,7 +515,7 @@ invoking ``make`` in the FreeRTOS BSP.
 
      .. code-block:: console
 
-       svn checkout https://github.com/MicroEJ/PlatformQualificationTools/trunk/framework/platform/content [path_to_platform_configuration_directory]
+       svn export --revision 54 --force https://github.com/MicroEJ/PlatformQualificationTools/trunk/framework/platform/content [path_to_output_directory]
```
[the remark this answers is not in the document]
microej-czaepfel: The point is that providing the revision number ensures repeatability, otherwise the latest revision is exported. The same applies to FreeRTOS , where the version is also specified.

microej-czaepfel:
> 
> 
> Thanks for the update.
> Comments inlined.
> 
> Also, you added an image `images/tuto_microej_fw_from_scratch_add_platform_configuration_additions.PNG ` but you didn't replace the previous one. Please do that.
> 
> In addition, be careful when adding files that the extension is in lower-case. Windows ignores the case but Linux don't so you might be committing a file with a `.PNG` and use `.png` in the rst. It will build fine on Windows but fail on Linux.
> To rename the file, you might need to do it in several steps.
> 
>     1. `mv foo.PNG bar.png && git commit`
> 
>     2. `mv bar.png foo.png && git commit --amend`

Are you sure about tuto_microej_fw_from_scratch_build_platform.PNG? It says file modified. Also the extension has been left unchanged, the files have been committed this way in the first place.

gjadi-mej:
> 
> 
> > Thanks for the update.
> > Comments inlined.
> > Also, you added an image `images/tuto_microej_fw_from_scratch_add_platform_configuration_additions.PNG ` but you didn't replace the previous one. Please do that.
> > In addition, be careful when adding files that the extension is in lower-case. Windows ignores the case but Linux don't so you might be committing a file with a `.PNG` and use `.png` in the rst. It will build fine on Windows but fail on Linux.
> > To rename the file, you might need to do it in several steps.
> > ```
> > 1. `mv foo.PNG bar.png && git commit`
> > 
> > 2. `mv bar.png foo.png && git commit --amend`
> > ```
> 
> Are you sure about tuto_microej_fw_from_scratch_build_platform.PNG? It says file modified. Also the extension has been left unchanged, the files have been committed this way in the first place.

No, you're correct.  Let it as is + with the new image you added.

Tutorials/tutorialCreateFirmwareFromScratch.rst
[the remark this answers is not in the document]
gjadi-mej: I prefer to keep using the latest released version and we will update the tuto from time to time.
I think I prefer to have the tutorial out of date rather than having users with out-of-date tools.

gjadi-mej:
Can you add "explicit" links to the platform configuration section and to the platform build section?
Something like: "For more information on <TOPIC>, please refer <link>".

gjadi-mej:
Regarding the links to the github's commit:
Is there a way to make them smaller?  Less intrusive?  Something akin to what you showed me with the "Download" small icon and link.  Except that, instead of downloading the patch, you go to github's commit.

Is that doable?  If not, maybe you can use the `caption` argument (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).

What do you think?

microej-czaepfel:
> 
> 
> Regarding the links to the github's commit:
> Is there a way to make them smaller? Less intrusive? Something akin to what you showed me with the "Download" small icon and link. Except that, instead of downloading the patch, you go to github's commit.
> 
> Is that doable? If not, maybe you can use the `caption` argument (https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).
> 
> What do you think?

I have used caption.

gjadi-mej (approved):

PR #260, opened 2021-06-08
Explain event consumption

gjadi-mej:
Provide guidelines about when to consume events and when not to consume events.

https://docs.microej.com/en/feature-m0137w-578_event_handling/ApplicationDeveloperGuide/UI/MWT/how-to-widget.html#consuming-events

### DIFF
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -89,7 +89,7 @@ The following guidelines are recommended to decide when to consume an event and 
    If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
    The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
    The user has to release or to drag the widget to trigger an action.
-   If the user presses a widget and then drags their finger or the pointer out of the widget before releasing it, the action is not triggered.
+   If the user presses a widget and then drags the pointer (e.g. their finger or a stylus) out of the widget before releasing it, the action is not triggered.
 
 Listening to the life-cycle hooks
 ---------------------------------
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -83,13 +83,13 @@ The following guidelines are recommended to decide when to consume an event and 
 
 - If the widget triggers an action when receiving the event, it consumes the event.
 - If the widget does not trigger an action when receiving the event, it does not consume the event.
-- If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
 
-  .. note::
+.. note::
 
-     The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
-     The user has to release it.
-     If the user presses a widget and then drags their finger out of the widget before releasing it, the action is not triggered.
+   If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
+   The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
+   The user has to release it.
+   If the user presses a widget and then drags their finger out of the widget before releasing it, the action is not triggered.
 
 Listening to the life-cycle hooks
 ---------------------------------
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -75,20 +75,15 @@ For example, the following snippet prints a message when the widget receives an 
 Consuming events
 ~~~~~~~~~~~~~~~~
 
-To indicate that an event was consumed by a widget, ``handleEvent()``
-should return ``true``.  When an event is consumed, it is not
-propagated to other widgets.  The widget that consumed the event, is
-the only one that can react to it.
+To indicate that an event was consumed by a widget, ``handleEvent()`` should return ``true``.
+When an event is consumed, it is not propagated to other widgets.
+The widget that consumed the event, is the only one that can react to it.
 
-The following guidelines are recommended to decide when to consume an
-event and when not to consume an event:
+The following guidelines are recommended to decide when to consume an event and when not to consume an event:
 
 - If the event triggers an action on the widget, consume the event.
-- If the event does not trigger an action the widget, do not consume
-  the event.
-- If the event is ``Pointer.PRESSED``, do not consume the event unless
-  it is required that the subsequent widgets in the hierarchy do not
-  receive it.
+- If the event does not trigger an action the widget, do not consume the event.
+- If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
 
 Listening to the life-cycle hooks
 ---------------------------------
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -77,7 +77,7 @@ Consuming events
 
 To indicate that an event was consumed by a widget, ``handleEvent()`` should return ``true``.
 Usually, once an event is consumed, it is not dispatched to other widgets (this behavior is controlled by the event dispatcher).
-The widget that consumed the event is the only one that can react to it.
+The widget that consumed the event is the last one to receive it.
 
 The following guidelines are recommended to decide when to consume an event and when not to consume an event:
 
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -89,7 +89,7 @@ The following guidelines are recommended to decide when to consume an event and 
    If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
    The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
    The user has to release it or drag it.
-   If the user presses a widget and then drags their finger out of the widget before releasing it, the action is not triggered.
+   If the user presses a widget and then drags their finger or the pointer out of the widget before releasing it, the action is not triggered.
 
 Listening to the life-cycle hooks
 ---------------------------------
--- a/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
+++ b/ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
@@ -76,14 +76,20 @@ Consuming events
 ~~~~~~~~~~~~~~~~
 
 To indicate that an event was consumed by a widget, ``handleEvent()`` should return ``true``.
-When an event is consumed, it is not propagated to other widgets.
-The widget that consumed the event, is the only one that can react to it.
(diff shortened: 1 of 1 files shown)


ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,24 @@ For example, the following snippet prints a message when the widget receives an
...
+
+To indicate that an event was consumed by a widget, ``handleEvent()``
+should return ``true``.  When an event is consumed, it is not
+propagated to other widgets.  The widget that consumed the event, is
+the only one that can react to it.
+
+The following guidelines are recommended to decide when to consume an
+event and when not to consume an event:
+
+- If the event triggers an action on the widget, consume the event.
```
seonmej, line 86: I don't like this formulation.
`If the widget triggers an action when receiving the event, it consumes the event.` maybe?
`Otherwise, it does not consume the event.`

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,24 @@ For example, the following snippet prints a message when the widget receives an
...
+- If the event does not trigger an action the widget, do not consume
+  the event.
+- If the event is ``Pointer.PRESSED``, do not consume the event unless
```
seonmej, line 89: I think that PRESSED not consuming the events is more a recommendation than a rule. We should explain why it is special: the user action is not finished, we don't know yet if the user will just click (press then release) on the widget or if he will drag, etc.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,24 @@ For example, the following snippet prints a message when the widget receives an
...
+To indicate that an event was consumed by a widget, ``handleEvent()``
+should return ``true``.  When an event is consumed, it is not
+propagated to other widgets.  The widget that consumed the event, is
+the only one that can react to it.
+
+The following guidelines are recommended to decide when to consume an
+event and when not to consume an event:
+
+- If the event triggers an action on the widget, consume the event.
+- If the event does not trigger an action the widget, do not consume
```
thomasponsmicroej, line 87:
```suggestion
- If the event does not trigger an action on the widget, do not consume
```

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,24 @@ For example, the following snippet prints a message when the widget receives an
 		return false;
 	}
 
+Consuming events
+~~~~~~~~~~~~~~~~
+
+To indicate that an event was consumed by a widget, ``handleEvent()``
+should return ``true``.  When an event is consumed, it is not
+propagated to other widgets.  The widget that consumed the event, is
```
seonmej, line 80:
```suggestion
propagated to other widgets.  The widget that consumed the event is
```

thomasponsmicroej, line 80:
```suggestion
dispatched to other widgets.  The widget that consumed the event, is
```

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,24 @@ For example, the following snippet prints a message when the widget receives an
 		return false;
 	}
 
+Consuming events
+~~~~~~~~~~~~~~~~
+
+To indicate that an event was consumed by a widget, ``handleEvent()``
+should return ``true``.  When an event is consumed, it is not
```
seonmej, line 79:
```suggestion
should return ``true``.  Usually when an event is consumed, it is not
```

thomasponsmicroej: @seonmej why usually? After an event is consumed it is not dispatched to other widgets right? Or do you mean that it depends on the event dispatcher where one can introduce any rule that he may want to apply ?

seonmej: Yes, "usually" because an EventDispatcher can choose to ignore the result of the handleEvent. (For example the EXITED event in the PointerEventDispatcher is sent to several widgets and containers without checking the result of handleEvent.)

thomasponsmicroej: So, shouldn't we say that the event dispatching "policy" depends on the EventDispatcher used by the application? 
The recommendations in this section will apply to the PointerEventDispatcher (default behavior) but maybe not with another dispatcher.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: ok

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: I used seon suggestion.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
seonmej: You are right, what we explain here applies to the PointerEventDispatcher.

But at the same time, some of the rules are good practices.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: I added a note, what do you think?

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,25 @@ For example, the following snippet prints a message when the widget receives an
...
+The following guidelines are recommended to decide when to consume an event and when not to consume an event:
+
+- If the widget triggers an action when receiving the event, it consumes the event.
+- If the widget does not trigger an action when receiving the event, it does not consume the event.
+
+.. note::
+
+   If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
+   The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
+   The user has to release it.
```
seonmej, line 91:
```suggestion
   The user has to release it or drag it.
```

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,25 @@ For example, the following snippet prints a message when the widget receives an
 		return false;
 	}
 
+Consuming events
+~~~~~~~~~~~~~~~~
+
+To indicate that an event was consumed by a widget, ``handleEvent()`` should return ``true``.
+Usually, once an event is consumed, it is not dispatched to other widgets (this behavior is controlled by the event dispatcher).
+The widget that consumed the event is the only one that can react to it.
```
apoudret-microej, line 80: not true, it is possible to react to an event without consuming an event (see Pointer.PRESSED handling in Button for example. other example is EXITED event which can be handled by every widget which received the PRESSED)

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,25 @@ For example, the following snippet prints a message when the widget receives an
...
+
+The following guidelines are recommended to decide when to consume an event and when not to consume an event:
+
+- If the widget triggers an action when receiving the event, it consumes the event.
```
apoudret-microej, line 84: this is not always desired (see examples above)

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
```diff
@@ -72,6 +72,25 @@ For example, the following snippet prints a message when the widget receives an
...
+- If the widget does not trigger an action when receiving the event, it does not consume the event.
+
+.. note::
+
+   If the event is ``Pointer.PRESSED``, do not consume the event unless it is required that the subsequent widgets in the hierarchy do not receive it.
+   The ``Pointer.PRESSED`` event is special because pressing a widget is usually not the deciding factor to trigger an action.
+   The user has to release it or drag it.
+   If the user presses a widget and then drags their finger out of the widget before releasing it, the action is not triggered.
```
apoudret-microej, line 92: remove "their finger" => "and then drags out of the widget"

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: Yes, but that's not what is written.  It says "when the event is consumed, the widget is the only one that can react to it".  In other words, when the widget is consumed, other widgets can't use it.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: There are special cases, but this are general guidelines.  Isn't that enough?

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: Hmm I find it less clear.  What don't you like about "their finger"?

seonmej: Pointer is not mandatorily a touch screen (but I agree it's 100% of our customers for now).

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
seonmej: I agree with Andy. Several widgets can "react" to the same event. The widgets that receive the event first can react to it without consuming it.
And I don't like the "react" verb. Maybe something like: `The widget that consumed the event is the last one to receive it.`

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
seonmej: I like it like that: we expose the guidelines then we present a special case where these guidelines can be violated.

apoudret-microej: ok

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
apoudret-microej: yes and on a touch screen you could use a stylus for example. maybe replace "their finger" by "the pointer"?

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: Oh you're right, I didn't consider this possibility. I'll fix this.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: So it could be a mouse pointer for example?
What about:
>     If the user presses a widget and then drags their finger (or cursor) out of the widget before releasing it, the action is not triggered.

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
gjadi-mej: fxied

apoudret-microej: ok for me

ApplicationDeveloperGuide/UI/MWT/how-to-widget.rst
apoudret-microej: imo we shouldn't make any reference to how to the user interacts physically with the screen. "pointer" is the word that is used in microui (regardless of whether it is controlled by a mouse, a stylus or a finger)

gjadi-mej: fxied

PR #270, opened 2021-06-18
M0078DDG-423 added warning block regarding cached build warnings

fpinel-microej:


README.rst
```diff
@@ -18,6 +18,11 @@ To perform a link check for all links and references::
 
     make linkcheck
 
+Warning
+~~~~~~~
+
+ReST syntax errors warnings are setup to make the build fail to avoid them being pushed to the repository unadressed. However, when building locally, resulting warnings are cached once source files have been read once (even during linkcheck) and won't reappear on second builds, making them falsely successful. These warnings aren't cached when building the public website on Read The Docs, but the current linkcheck call output doesn't allow to know what caused the failure. To have them reappear when building locally, you can clear the `_build` folder, or checkout to another Git branch.
```
frivieremicroej, line 24:
```suggestion
Syntax warnings are configured to make the build fail. When building locally, warnings are cached once source files have been read once (even during linkcheck) and thus won't reappear on subsequent builds. 
To get all the warnings locally, clean the `_build` directory (``make clean``) and rebuild the whole documentation (``make html`` , ``make linkcheck``).
```

README.rst
```diff
@@ -18,6 +18,11 @@ To perform a link check for all links and references::
 
     make linkcheck
 
+Warning
```
frivieremicroej, line 21:
```suggestion
Warnings Processing
```

frivieremicroej (changes requested):
I made some suggestions to simplify the wording

README.rst
fpinel-microej: Actually this "title" was supposed to be not a title but a "Warning !" (I even wanted to use a ReST warning:: directive to make a visible red block but GitHub doesn't seem to support that so the point is moot) 
if you're fine with this as a sub-title it's fine but I hope it is visible and "catchy" enough for people to actually read it.

README.rst
fpinel-microej: Changed.

PR #273, opened 2021-06-22
M0078DDG-417 Document available IoT libraries

pauffray:


frivieremicroej (approved):

ApplicationDeveloperGuide/networking.rst
```diff
@@ -0,0 +1,183 @@
...
+
+   *  -  Name
+      -  Description
+      -  Module Link
+      -  Use
+   
+   *  -  ``Bluetooth``
+      -  Bluetooth Low Energy (BLE) based on the Generic Attribute Profile (GATT).
+      -  `bluetooth <https://repository.microej.com/modules/ej/api/bluetooth/>`_
+      -  - `Bluetooth utility Library <https://repository.microej.com/modules/ej/library/iot/bluetooth-util/>`_
```
tdelhomenie-microej, line 32: As a developer I would like a description of the API features, the Ivy dependency snippet to quickly setup my project, and documentation on how to use it:
- API features: the description is very light, it is hard to know what the API contains. Don't we have more detailed description ? A link to a README at least ?
- Ivy dependency: the link in the "Module link" column allows to find the information about the module but it is not easy, especially for someone not familiar with Ivy. We should add the Ivy coordinates directly to make the developer life easier.
- documentation: we need documentation about API usage. Examples are a good thing but it is not easy for developers to extract what they want. And some library do not even have example. Don't we have documentation for these APIs ? Or at least a javadoc link ?

ApplicationDeveloperGuide/networking.rst
```diff
@@ -0,0 +1,183 @@
+.. _networking:
+
+Networking
+==========
+
+This section presents networking libraries.
+
+The following schema shows the overall architecture and modules:
+
+.. figure:: images/networking_overview.png
```
tdelhomenie-microej, line 10: It should be "Example Applications" instead of "Examples Applications" IMO.

ApplicationDeveloperGuide/networking.rst
```diff
@@ -0,0 +1,183 @@
...
+      - 
+
+   *  -  ``SSL``
+      -  Client and Server secure sockets layer using Transport Layer Security (TLS) protocols.
+      -  `ssl <https://repository.microej.com/modules/ej/api/ssl/>`_
+      -  -  `SSL mutual client Example <https://github.com/MicroEJ/Example-Sandboxed-IOT/tree/master/com.microej.example.iot.ssl.mutual>`_
+         -  `SSL mutual server Example <https://github.com/MicroEJ/Example-Sandboxed-IOT/tree/master/com.microej.example.iot.ssl.mutual.server>`_
+         -  `SSL utility Library <https://repository.microej.com/modules/ej/library/iot/ssl-util/>`_
+
+IoT Libraries
```
tdelhomenie-microej, line 63: Maybe we should put these chapters in a "Add-On Libraries" chapter to match with the schema ?

tdelhomenie-microej (changes requested):

ApplicationDeveloperGuide/networking.rst
frivieremicroej: - API Features: the goal of this first page is to start to list libraries that have never been documented. This is a first version we could enhance in the future when we will have more documentation. Please note that a link to the README has been added when it contains valuable information.
- Ivy Dependency: I thought it too, but I finally did not do it for 2 reasons: 
  - 1) it avoids to put lots of hardcoded versions (the link points to the module name, not a specific version); 
  - 2) the dependency line would be too large to fit in a table column.
- The `Use` column contains what is publicly available (examples, README, user manual or other utility libraries). Sure we need many more documentation for each library in the future. Again this is not the purpose of this PR to produce the library documentation, but just a summary of what is already available.

frivieremicroej: For Javadoc, please have a look to feature/M0078DDG-417_documentation_library_iot_with_javadoc_link
I started a new column for Foundation Libraries, what do you think about that ?

ApplicationDeveloperGuide/networking.rst
frivieremicroej: done

tdelhomenie-microej (approved):

PR #283, opened 2021-07-06
M0078DDG-426 set permalinks back up on page titles

fpinel-microej:


conf.py
```diff
@@ -52,7 +52,7 @@
 html_show_sphinx = False
 
 # remove headers permalinks
```
frivieremicroej, line 54:
```suggestion
# # remove headers permalinks
```

frivieremicroej (approved):

PR #294, opened 2021-08-12
Feature/m0090 ide 3636 cc sources default value

seonmej:


ApplicationDeveloperGuide/moduleNatures.rst
```diff
@@ -508,6 +508,9 @@ This plugin defines the following configuration properties:
    * - microej.testsuite.properties.s3.cc.activated
      - When this property is set to true, the code coverage analysis is enabled.
      - ``true``
+   * - cc.src.folders
```
frivieremicroej, line 511: shall we put a note "MicroEJ SDK 5.4.0 or higher" ?

tdelhomenie-microej: The property `cc.src.folders` is available since 5.4.1, but the default value is changed in 5.5.0. We can add notes for both. And we can either wait the release of 5.5.0 to merge this PR, or split the PR to merge the addition of the property now and create another one for the default value for 5.5.0.

PR #295, opened 2021-08-13
M0078DDG-433 Update CI tutorial to use Artifactory 7

tdelhomenie-microej:
Use Artifactory 7 in the CI tutorial since the OSS version of Artifactory  6 is not available anymore for Windows.
Main changes are:
- Menus have changed
- Repositories `libs-snapshot-local` and `libs-release-local` are not created by default anymore
- New repositories are now accessible to anyone by default

Doc preview: https://docs.microej.com/en/feature-m0078ddg-433_artifactory-7/


tdelhomenie-microej:
@frivieremicroej @fpinel-microej the repositories libs-snapshot-local and libs-release-local are not created by default anymore in Artifactory 7. In the tuto we don't mention the repository libs-release-local but we reference it in the ivy settings file. This is a problem now that is not created by default anymore. We should either remove it from the ivy settings (but it means there is no release settings), or create it in the tuto whereas we don't use it. Any opinion ?

frivieremicroej:
> @frivieremicroej @fpinel-microej the repositories libs-snapshot-local and libs-release-local are not created by default anymore in Artifactory 7. In the tuto we don't mention the repository libs-release-local but we reference it in the ivy settings file. This is a problem now that is not created by default anymore. We should either remove it from the ivy settings (but it means there is no release settings), or create it in the tuto whereas we don't use it. Any opinion ?

Even if the tutorial does not explicitly indicates how to make a release, I think it is better to keep the settings file correctly configured for publishing a module in release mode.

Now we are free to rename them: `custom-modules-snapshot` and `custom-module-release`

frivieremicroej:
For the release repository, please also add the option to configure it to reject to write a module that already here (immutable) 
 ask to @asogno-mej if necessary. (when creating new repositories, we often miss this option)

tdelhomenie-microej:
> Even if the tutorial does not explicitly indicates how to make a release, I think it is better to keep the settings file correctly configured for publishing a module in release mode.
Now we are free to rename them: custom-modules-snapshot and custom-module-release

Done.

> For the release repository, please also add the option to configure it to reject to write a module that already here (immutable)
ask to @asogno-mej if necessary. (when creating new repositories, we often miss this option)

The permissions for Delete/Overwrite on repositories is not enabled so users cannot do it.

PR #315, opened 2021-09-28
document "[M25] Assertion code cannot be removed..."

frivieremicroej:


akhabali:
Can we add more details showing which options can be used to deactivate the remove assert optimization. 
It may helps working around the issue but I'm not sure if it's the suitable place for such info

akhabali (approved):
LGTM

PR #320, opened 2021-09-30
document known issue with if constant at the end of try block

frivieremicroej:
This issue exists from the beginning of BON Constant introduction
(7.11.0), but may throw different errors depending on the upper
architecture version


akhabali:
LGTM

akhabali (approved):

PR #325, opened 2021-10-18
update with glossary case sensitive terms

frivieremicroej:


frivieremicroej:
Also @emoreac-microej , I would like to replace "platform watchdog timer" by a term which is not MicroEJ Platform specific: "hardware watchdog timer" ? or "system watchdog timer" ?

emoreac-microej:
> Also @emoreac-microej , I would like to replace "platform watchdog timer" by a term which is not MicroEJ Platform specific: "hardware watchdog timer" ? or "system watchdog timer" ?

Hi @frivieremicroej ,
You can use the term "hardware watchdog timer". This one is more accurate because the purpose of this foundation library is to manipulate the board's hardware watchdog timer.

PR #329, opened 2021-10-25
Feature/tutorial null annotations java 8 and up

tboue:
Tutorial on how to use null annotations from java 8 and up. Also explains how to use external annotations.


tboue:
Application developers won't work with JAVA 8+, split doc https://github.com/MicroEJ/docs/pull/335 and the part on JAVA 8+ only moved to private documentation.

PR #369, opened 2022-02-17
fix usage of resident application

frivieremicroej:
- indicate the previous name in glossary as it was widely used in the
past, especially in samples README and forum posts
- replace Resident application -> System application except in plugin
label (Firmware Linker)


akhabali (approved):
LGTM

PR #376, opened 2022-03-07
document license check

frivieremicroej:


SDKUserGuide/licenses.rst
```diff
@@ -27,6 +27,25 @@ The list of installed licenses is available in the SDK preferences dialog page i
...
+
+   * - **Application**
+     - **Run on Simulator**
+     - **Build on Device**
+   * - Standalone Application
+     - NO
+     - **YES**
+   * - Sandboxed Application
+     - NO
+     - NO
```
tdelhomenie-microej, line 47: Shouldn't we add the build of a Kernel ? You want to stick with Applications only ?

frivieremicroej: image

OK we can add it to clarify

tdelhomenie-microej (approved):

PR #410, opened 2022-06-03
Update the default value for the test include pattern to match the latest version of abstract-test plugin

gjadi-mej:
Wait until the module is released.


gjadi-mej:
Closing, this is not the solution kept.

PR #411, opened 2022-06-03
M0090IDE-3814 Change default targets of MMM CLI

tdelhomenie-microej:
Change default targets of MMM CLI to `clean,verify`.


denisbedin (approved):
LGTM

PR #414, opened 2022-06-09
M0137W-687 Clarify NLS External Loader usage

tdelhomenie-microej:


thomasponsmicroej (approved):
what about putting the NLS External Loader under ADG/Application resources?

PR #428, opened 2022-07-25
M0078DDG-486 Document SNI limitations

jerome-leroux:


ApplicationDeveloperGuide/limitations.rst
```diff
@@ -40,6 +40,9 @@ Please consult :ref:`architecture_changelog` for limitations changes on former v
    * - Numbers of exception handlers per method
      - 63
      - 63
+   * - Number of paramaters for an SNI method
```
rlouedec-microej, line 43: Typo:
```suggestion
   * - Number of parameters for an SNI method
```

rlouedec-microej (changes requested):
Except typo, lgtm.

Another general comment: does it make sense to keep the eval/prod columns ?
I would rather split the table in two: 1 with general limitation, 1 with eval/prod differences.
But I'm fine with one table.

ApplicationDeveloperGuide/limitations.rst
jerome-leroux: Yes, we could review the table structure. There are not so much differences between eval and prod now.
This could be done later.

rlouedec-microej (approved):

PR #430, opened 2022-08-11
Add a dependency example for UI pack

balix-microej:
The [installation section ](https://docs.microej.com/en/latest/PlatformDeveloperGuide/uiMicroui.html#installation) states that the library is an additional module and can be enabled via the platform configuration editor.

However it doesn't explain that the UI pack dependency should be added to the platform `module.ivy` file beforehand.

### DIFF
--- a/PlatformDeveloperGuide/uiMicroui.rst
+++ b/PlatformDeveloperGuide/uiMicroui.rst
@@ -154,8 +154,17 @@ is defined during the pre-generation of a font (see
 Installation
 ============
 
-The MicroUI library is an additional module. In the platform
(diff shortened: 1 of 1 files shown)


gbalan-microej:
I understand your changes but I think it is "correct":
- not coherent with network for instance: https://docs.microej.com/en/latest/PlatformDeveloperGuide/networkCoreEngine.html#installation
- MicroUI is a part of UI pack, like Display, Input etc. May be a dedicated chapter that describes the UI pack format and available modules is required. And the chapters "installation" of each module can refer to this chapter.

frivieremicroej:
Unfortunately, this PR is inactive for a while. Please reopen if someone wants to resume.

PR #444, opened 2022-09-13
M0078DDG-483 Add information on Android String resource in NLS docume…

apoudret-microej:
…ntation

### DIFF
--- a/ApplicationDeveloperGuide/nls.rst
+++ b/ApplicationDeveloperGuide/nls.rst
@@ -3,24 +3,23 @@
 Native Language Support
 =======================
 
-Overview
---------
+Introduction
+------------
 
 Native Language Support (NLS) allows the application to facilitate internationalization.
 It provides support to manipulate messages and translate them in different languages.
 Each message to be internationalized is referenced by a key, which can be used in the application code instead of using the message directly.
 
-Messages must be defined in `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
-Here is an example:
+Localization Source Files
+-------------------------
+
+Messages must be defined in localization source files, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
+
+Localization source files can be either `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
+
+Here is an example of a PO file:
 
 ::
-
-   msgid ""
-   msgstr ""
-   "Language: en_US\n"
-   "Language-Team: English\n"
-   "MIME-Version: 1.0\n"
-   "Content-Type: text/plain; charset=UTF-8\n"
 
    msgid "Label1"
    msgstr "My label 1"
@@ -28,7 +27,24 @@ Here is an example:
    msgid "Label2"
    msgstr "My label 2"
 
-PO files are declared in :ref:`Classpath<chapter.microej.classpath>` ``*.nls.list`` files (**and** to ``*.externresources.list`` for an external resource, see :ref:`chapter.microej.applicationResources`).
+And here is an example of an Android String resource:
+
+.. code-block:: xml
+
+   <resources>
+      <string name="Label1">My label 1</string>
+      <string name="Label2">My label 2</string>
+   </resources>
+
+.. note::
+
+   When using Android String resources, `string arrays <https://developer.android.com/guide/topics/resources/string-resource#StringArray>`_ are also supported.
+   However, `plurals <https://developer.android.com/guide/topics/resources/string-resource#Plurals>`_ are not supported.
+
+NLS List Files
+--------------
+
+Localization source files are declared in :ref:`Classpath<chapter.microej.classpath>` ``*.nls.list`` files (**and** to ``*.externresources.list`` for an external resource, see :ref:`chapter.microej.applicationResources`).
 
 .. graphviz::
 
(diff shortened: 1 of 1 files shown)


ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,46 @@
...
+------------
 
 Native Language Support (NLS) allows the application to facilitate internationalization.
 It provides support to manipulate messages and translate them in different languages.
 Each message to be internationalized is referenced by a key, which can be used in the application code instead of using the message directly.
 
-Messages must be defined in `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
-Here is an example:
+Definition
+----------
```
tdelhomenie-microej, line 14: Not sure it is worth adding another level of title here, I would keep `Localization Source Files` and `NLS List Files` at the first level. WDYT ?

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,46 @@
...
+Messages must be defined in localization source files, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
+
+Localization source files can be `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
 
-   msgid ""
-   msgstr ""
-   "Language: en_US\n"
-   "Language-Team: English\n"
-   "MIME-Version: 1.0\n"
-   "Content-Type: text/plain; charset=UTF-8\n"
```
tdelhomenie-microej, line 23: Why removing the header of the PO file? The goal of this page is also to allow the developer to quickly have a working example by copy/paste.

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,46 @@
...
+Definition
+----------
 
-::
+Localization Source Files
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Messages must be defined in localization source files, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
+
+Localization source files can be `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
```
tdelhomenie-microej, line 21: Do we support all the feature of Android String resources? I had in mind we don't but I might be wrong. If we don't we should document the limitations.

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,46 @@
...
+::
 
    msgid "Label1"
    msgstr "My label 1"
 
    msgid "Label2"
    msgstr "My label 2"
 
-PO files are declared in :ref:`Classpath<chapter.microej.classpath>` ``*.nls.list`` files (**and** to ``*.externresources.list`` for an external resource, see :ref:`chapter.microej.applicationResources`).
+And here is an example of Android String resource:
```
tdelhomenie-microej, line 33:
```suggestion
And here is an example of a Android String resource:
```

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,46 @@
...
+
+Localization source files can be `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
 
-   msgid ""
-   msgstr ""
-   "Language: en_US\n"
-   "Language-Team: English\n"
-   "MIME-Version: 1.0\n"
-   "Content-Type: text/plain; charset=UTF-8\n"
+Here is an example of PO file:
```
tdelhomenie-microej, line 23:
```suggestion
Here is an example of a PO file:
```

tdelhomenie-microej (changes requested):
@apoudret-microej The build is failing due to "Title level inconsistent" warnings : https://readthedocs.com/projects/microej-docs/builds/1095280/

ApplicationDeveloperGuide/nls.rst
apoudret-microej: agree, fixed

ApplicationDeveloperGuide/nls.rst
apoudret-microej: it is a working example (header isn't necessary)

ApplicationDeveloperGuide/nls.rst
apoudret-microej: fixed

ApplicationDeveloperGuide/nls.rst
apoudret-microej: correct. i added a note under the Android string resource example

tdelhomenie-microej (approved):

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,48 @@
...
 It provides support to manipulate messages and translate them in different languages.
 Each message to be internationalized is referenced by a key, which can be used in the application code instead of using the message directly.
 
-Messages must be defined in `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
-Here is an example:
+Localization Source Files
+-------------------------
 
-::
+Messages must be defined in localization source files, located in the Classpath of the application (for example in the ``src/main/resources`` folder).
```
frivieremicroej, line 16: why for example ? is there an other way ?
```suggestion
Messages must be defined in localization source files, located in the Classpath of the application (in the ``src/main/resources`` folder).
```

ApplicationDeveloperGuide/nls.rst
```diff
@@ -3,32 +3,48 @@
...
+
+Localization source files can be `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
```
frivieremicroej, line 18:
```suggestion
Localization source files can be either `PO files <https://www.gnu.org/software/gettext/manual/gettext.html#PO-Files>`_ or `Android String resources <https://developer.android.com/guide/topics/resources/string-resource>`_.
```

frivieremicroej: we should explicitely indicate this can be a mix of both (right ?) is there any limitation/conflict ?

ApplicationDeveloperGuide/nls.rst
```diff
@@ -52,56 +68,36 @@ Example:
...
-For each line, PO files whose name starts with the interface name (``Messages`` and ``Labels``
-in the example) are retrieved from the Classpath and used to generate:
-
-- a Java interface with the given FQN, containing a field for each ``msgid`` of the PO files
-- a NLS binary file containing the translations
-
-So, in the example, the generated interface ``com.mycompany.myapp.Labels`` will gather all the 
-translations from files named ``Labels*.po`` and located in the Classpath.
-PO files are generally suffixed by their locale (``Labels_en_US.po``) but it is only for convenience
-since the suffix is not used, the locale is extracted from the PO file's metadata.
```
frivieremicroej, line 82: the locale is extracted from the PO file's metadata => did you change the behavior of PO files when adding Android Resources ?

ApplicationDeveloperGuide/nls.rst
```diff
@@ -52,56 +68,36 @@ Example:
...
+This module includes an Add-On Processor which parses the localization source files.
+For each interface declared in the NLS list files, all the localization source files whose names start with the interface name are used to generate:
 
-      String label = Labels.NLS.getMessage(Labels.Label1);
+- a Java interface with the given FQN, containing a field for each message of the localization source files
+- a NLS binary file containing the translations
 
-      ...
+So, in the example, the generated interface ``com.mycompany.myapp.Labels`` will gather all the 
+translations from files named ``Labels*`` and located in the Classpath.
```
frivieremicroej, line 89: Please clarify the resource name: Do we search for "com/mycompany/myapp/Labels*" or "**/Labels*" resources ?

frivieremicroej (changes requested):

ApplicationDeveloperGuide/nls.rst
apoudret-microej: the name of the locale is defined like this:
- if it is a PO file, we look at the metadata. if there is no metadata, we use the suffix in the filename
- if it is an Android string resource, we always use the suffix in the filename

we discussed it with Thomas this morning. we agreed that it is simpler if we say in the documentation that the filename should always have the locale name as suffix. it seems like a good practice anyway

ApplicationDeveloperGuide/nls.rst
apoudret-microej: i don't know if both can be used for the same interface.. but i don't think someone will ever do that

ApplicationDeveloperGuide/nls.rst
apoudret-microej: they can also be in a lib that the app depends on

ApplicationDeveloperGuide/nls.rst
apoudret-microej: i changed to:

`from files named ``Labels*`` and located in any package of the Classpath`

ApplicationDeveloperGuide/nls.rst
frivieremicroej: ok... but to build the lib, they must put the resource in the ``src/main/resources`` of the lib project, so there is no other way.

frivieremicroej: ok for ie

ApplicationDeveloperGuide/nls.rst
frivieremicroej: OK so either/or

ApplicationDeveloperGuide/nls.rst
frivieremicroej: OK, you could check that the header matches the suffix to avoid unwanted behaviour

frivieremicroej (approved):

PR #450, opened 2022-09-20
M0078DDG-316 Javadoc references in tutorials section

fpinel-microej:


fpinel-microej:
Scroll List page (step11) of the Get Started With GUI section mention a `Scroll` container class, not sure what this is from or if it is supposed to represent a user-implemented container.

frivieremicroej (approved):

PR #503, opened 2023-01-20
Fix options for "No BSP Connection using SDK"

frivieremicroej:
According to documentation "deploy.bsp.microejscript=true" requires Platform at least configured with Partial BSP Connection

Setting 'deploy.dir.microejscript' is sufficient and triggers the build script (just remove the option if you do not want to trigger the build script)


thomasdelhomenie (approved):
Tested successfully! 👍

tdelhomenie-microej (approved):

PR #509, opened 2023-01-26
Feature/m0090 de 4108 dropins doc status updating

levMicroej:


SDK6UserGuide/limitations.rst
```diff
@@ -8,6 +8,7 @@ The SDK 6 allows to:
 - build the Java artifact of Applications and Add-On Libraries.
 - execute the tests of a project with the Simulator.
 - execute the Artifact Checker on a project.
+- load of the VEE Port from the dropins folder.
```
tdelhomenie-microej, line 11: if we add this line in the capabilities of the SDK, we should also add the other ways to load a VEE Port. So I would move this line after the line "run Applications with the Simulator." and enrich it like that:
```
- load the VEE Port from its archive file path, its folder path, a dependency or from the ``dropins`` folder.
```

SDK6UserGuide/selectVeePort.rst
```diff
@@ -50,18 +50,26 @@ There are 3 different ways to provide a VEE Port in the build file of the projec
       
       Refer to `the Gradle documentation <https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.dsl.DependencyHandler.html>`__ 
       to learn all the options to select dependencies.
+      
+- Copy a VEE port file to the dropins folder. The default dropins folder location is ``[module_project_dir]/dropins``. It can be changed using the build property ``veePortDropinsDir``::
```
tdelhomenie-microej, line 54:
```suggestion
- Copy a VEE port archive file to the ``dropins`` folder. The default dropins folder location is ``[module_project_dir]/dropins``. It can be changed using the build property ``veePortDropinsDir``::
```

tdelhomenie-microej (changes requested):

tdelhomenie-microej (approved):

PR #511, opened 2023-02-13
add new ARMv7A and ARMv7VE supported architectures

agazardMicroeJ:


frivieremicroej:
@agazardMicroeJ the build fails. Also, is it planned to build these Architectures for versions 7.x or only 8.x (if this latter case, we will change the target branch)

agazardMicroeJ:
@frivieremicroej It is only planned for Architectures 8.x. Which target branch should I use?

jerome-leroux (approved):
I removed the link to the modules because they are not published yet and it broke the doc build.

PR #518, opened 2023-02-27
Update null analysis section for edc version requirement

emoreac-microej:


ApplicationDeveloperGuide/nullAnalysis.rst
```diff
@@ -122,6 +129,21 @@ Advanced Use
...
+
+
+Troubleshooting
+---------------
+
+The project cannot build anymore after Null Analysis setup
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. figure:: images/null_analysis_troubleshooting_edc_version.png
+  :alt: Null Analysis wrong edc version
```
tdelhomenie-microej, line 141: We should add the exception with the full stacktrace as text also, so it will be searchable.

tdelhomenie-microej (changes requested):

tdelhomenie-microej (approved):

PR #533, opened 2023-03-20
document how to skip a Gradle task

denisbedin:
https://docs.microej.com/en/feature-microej-sdk-6-skip-task/SDK6UserGuide/howtos/gradleSkipTask.html

### DIFF
--- a/SDK6UserGuide/howtos/gradleSkipTask.rst
+++ b/SDK6UserGuide/howtos/gradleSkipTask.rst
@@ -1,0 +1,60 @@
+.. _sdk_6_howto_gradle_skip_task:
+
+How To Skip a Gradle Task
+=========================
+
+When a task is executed, it is possible to skip one or more of the tasks on which the called task depends.
+For example, you can skip the ``test`` task if you want to build a project without executing the tests.
+
+If you want to skip a task, one of the following ways can be used :
+
+- Add the ``-x`` or ``--exclude-task`` option in the command line::
+
+   gradle build -x test
+
+  The task is skipped for this execution only.
+
+- Exclude the task in the build script of the project ::
+
+   project.gradle.startParameter.excludedTaskNames.add("test")
+
+  The task is never executed.
+
+When one of these two ways is used, not only the task but also all the tasks on which it depends are skipped.
+For example, if you choose to skip the ``test`` task, all the tasks which are used to produce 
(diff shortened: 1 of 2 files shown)


SDK6UserGuide/howtos/gradleSkipTask.rst
```diff
@@ -0,0 +1,51 @@
...
+
+   gradle build -x test
+
+  The task is skipped for this execution only.
+
+- Exclude the task in the build script of the project ::
+
+   project.gradle.startParameter.excludedTaskNames.add("test")
+
+  The task is never executed.
```
tdelhomenie-microej, line 21: Isn't there a difference between excluding a task with `-x` and with `project.gradle.startParameter.excludedTaskNames`. I had in mind that when you use `-x` it does exclude the tasks on which the excluded task is dependent. Am I wrong?

tdelhomenie-microej (changes requested):

SDK6UserGuide/howtos/gradleSkipTask.rst
denisbedin: I didn't notice this difference between the 2 ways to exclude a task. I tried with the `test` task, that depends on the `testClasses` task, and if I exclude the `test` task, the `testClasses` task is also excluded regardless of the way used to exclude the task (`-x` or  `project.gradle.startParameter.excludedTaskNames`).

tdelhomenie-microej: Right, I don't remember I had this assertion in mind.

tdelhomenie-microej (approved):

SDK6UserGuide/howtos/gradleSkipTask.rst
```diff
@@ -0,0 +1,62 @@
...
+
+When one of these two ways is used, not only the task but also all the tasks on which it depends are skipped.
+For example, if you choose to skip the ``test`` task, all the tasks which are used to produce 
+the test runtime classpath are also skipped.
+
+The next section describes how you can skip a task, but still execute the tasks on which it depends.
```
tdelhomenie-microej, line 27: I would remove this line, it is redundant with the next one.

denisbedin: Done.

PR #536, opened 2023-03-22
Update SDK 6 to version 0.5.0 and set the Gradle compatible version to 8.0.2 and higher

denisbedin:
https://docs.microej.com/en/feature-microej-sdk-6-use-gradle-8.x/

### DIFF
--- a/SDK6UserGuide/install.rst
+++ b/SDK6UserGuide/install.rst
@@ -51,7 +51,7 @@ Install Gradle
 --------------
 
 Once a JDK is correctly configured, the next step is to install Gradle by following `the official documentation <https://gradle.org/install/>`__.
-The SDK is only compatible with the Gradle ``7.x``, so make sure to install a right version.
+The SDK is only compatible with the Gradle ``8.0.2`` and higher, so make sure to install a right version.
 Once done, you can verify your installation by opening a terminal and run the command ``gradle -v``.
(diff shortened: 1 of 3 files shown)


SDK6UserGuide/install.rst
```diff
@@ -51,26 +51,26 @@ Install Gradle
 --------------
 
 Once a JDK is correctly configured, the next step is to install Gradle by following `the official documentation <https://gradle.org/install/>`__.
-The SDK is only compatible with the Gradle ``7.x``, so make sure to install a right version.
+The SDK is only compatible with the Gradle ``8.x``, so make sure to install a right version.
```
tdelhomenie-microej, line 54: It is 8.0.2 minimum actually, so I would suggest:
```suggestion
The SDK is only compatible with the Gradle ``8.0.2`` and higher, so make sure to install a right version.
```
or
```suggestion
The SDK is only compatible with the Gradle ``8.x`` (at least ``8.0.2``), so make sure to install a right version.
```

tdelhomenie-microej (changes requested):

SDK6UserGuide/install.rst
denisbedin: yes, I chose the first suggested change.

tdelhomenie-microej (approved):

PR #567, opened 2023-05-10
M0081MEJA-2070 update the changelog.

dalgassimou:


frivieremicroej:
This is related to SOARv2 => no change from the user point of view

PR #584, opened 2023-05-24
document the trace analyzer activity timeout

frivieremicroej:


tboue (approved):
Tested with value 7 : in PQT like runner got :
`[testsuite:javaTestsuite] [traceAnalyzer:socketTraceAnalyzer] 1 : TraceAnalyzer ERROR :
[testsuite:javaTestsuite] [traceAnalyzer:socketTraceAnalyzer] [M3] - No activity on input stream since 7 s.`

PR #626, opened 2023-06-07
Add instruction to use a license dongle with WSL.

bguedas:


frivieremicroej:
@fpinel-microej please review by testing and running the instrutions on your side

PR #634, opened 2023-06-13
Mention resolve foundation library in workspace in mock page

gjadi-mej:


VEEPortingGuide/mock.rst
```diff
@@ -289,6 +289,8 @@ Once the module is built, the mock can be installed in a Platform in one of the
 - by adding the mock module as a regular Platform :ref:`module dependency <mmm_module_dependencies>` (if your Platform configuration project contains a ``module.ivy`` file), 
 - or by manually copying the JAR file ``[mock_project]\target~\rip\mocks\[mock_name].jar`` to the :ref:`Platform configuration <platform_configuration_creation>` mock dropins folder ``dropins/mocks/dropins/``.
 
+During development, enable :ref:`resolve_foundation_libraries_in_workspace` to use the mock without having to install it after each modification.
```
tdelhomenie-microej, line 292: I would use "make sure that ..." instead of "enable" since it is supposed to be enabled by default.

gjadi-mej: How come? I've Dist 21.11 with SDK 5.7.0 and it was turned off. Clicking restore default does not select it. (but it does not unselect it either).

tdelhomenie-microej: You mean in a new workspace? It should not.

gjadi-mej: I did a test with a fresh workspace and it was enabled. I don't know why it wasn't in my previous test.

The "restore default" doesn't really work as I expect, I'll open a separate ticket for that.
I'll update this MR according to your initial suggestion.

tdelhomenie-microej (approved):

PR #648, opened 2023-06-29
Feature/m0092 meja UI 2874 nemagfx

gbalan-microej:
cf [doc not built because the CCO is not available on developer repo]


gbalan-microej:
replaced by #652

PR #665, opened 2023-07-24
 Reformulate a bit as VEE Ports are imported in SDK6

gjadi-mej:
Saying "import a VEE Port" there caused me confusion because I was used to importing VEE Ports in my SDK (in SDK5).
But that is not the workflow in SDK6.
Remove the word "import" because the VEE Ports are not imported, we merely point at them.


denisbedin (approved):
LGTM but you must rebase your branch on the feature/microej-sdk-6 branch

PR #666, opened 2023-07-24
Feature/sdk6 misc docs improvements

gjadi-mej:


SDK6UserGuide/index.rst
```diff
@@ -48,13 +48,13 @@ The SDK is licensed under the :ref:`SDK End User License Agreement (EULA) <sdk_e
    createProject
    importProject
    buildProject
+   selectVeePort
```
tdelhomenie-microej, line 51: The initial idea was to start the documentation by the most common actions a developer would want to do, so create/import a project, build it and run it (sort of like in a tutorial). That's why the selection of a VEE Port has been added after, it is kind of a sub-part of one of the main topics. WDYT?

tdelhomenie-microej (approved):

PR #668, opened 2023-08-02
Add SDK Distribution max version for JavaSoft (Oracle) registry keys option

tdelhomenie-microej:
Add SDK Distribution max version for JavaSoft (Oracle) registry keys option<


rlouedec-microej (approved):
LGTM

PR #672, opened 2023-08-10
Explain how translations are loaded by binary-NLS

rlouedec-microej:


ApplicationDeveloperGuide/nls.rst
```diff
@@ -10,6 +10,36 @@ Native Language Support (NLS) allows the application to facilitate international
...
+Principle
+---------
+
+NLS is distributed as an add-on library containing a single Java interface: `NLS`_.
+
+In addition to that, the binary-nls library provides a factory for implementations of this interface:
+it uses an :ref:`add-on processor <module_natures.addon_processor>` which processes, offboard, the
+Localization Source Files into one `BON resource buffer`_ file for compactness.
+
+During the :ref:`clinit <soar_clinit>` phase, this resource file is opened and the list of locales is parsed.
```
apoudret-microej, line 22: i suggest to move/merge this paragraph into "BinaryNLS Resource Generation" section

ApplicationDeveloperGuide/nls.rst
```diff
@@ -10,6 +10,36 @@ Native Language Support (NLS) allows the application to facilitate international
...
+After that, the resource remains opened for the rest of the Application execution and is directly used to
+retrieve messages translations for the supported locales.
+
+Note that this implies that the binary-nls implementation does not support:
```
apoudret-microej, line 26: suggest to move the lines from "Note that this implies" until "the added/modified locale)." to a dedicated section ("Limitations" or "Dynamic Modifications") above "NLS External Loader" section

ApplicationDeveloperGuide/nls.rst
```diff
@@ -10,6 +10,36 @@ Native Language Support (NLS) allows the application to facilitate international
...
+
+* to dynamicly add a new locale
+* to dynamicly modify messages translations
+
+even when the resource is external (see :ref:`External resource loader <section_externalresourceloader>`).
+Note that binary-nls does not protect against dynamic update of the resource which may lead to invalid
```
apoudret-microej, line 32: i think that this is relevant for all kinds of resources: external resources must not be removed/changed while they are open by the app. for example one should not remove an external image if it is open by an app. maybe we can add/move this info into external resources loader doc?

ApplicationDeveloperGuide/nls.rst
rlouedec-microej: I understand that they are related but I am not sure we should merge those.

My idea is that a "software architect" should only need to read the introduction and principle section.
Following sections are for developers that actually implement a solution with NLS.

Still, I added a link from this section to the principle section and clarify a few points in ad88bed2199a39bae7a81cfe7ff1888e19de4343
WDYT ?

ApplicationDeveloperGuide/nls.rst
rlouedec-microej: Good point, this will clarify that this is potentially subject to later improvements.

For the same reason discussed in https://github.com/MicroEJ/docs/pull/672#discussion_r1291632840, I am not about the location of this section but ok, I am fine with your suggestion.

I also took the opportunity to clarify another limitation you reported (no way to close the resource).

Done in: 82a92a2b35dba989d96a3968500fdd61303bc9b9

ApplicationDeveloperGuide/nls.rst
rlouedec-microej: I am not so sure about that: this depends on the consumer of the external resource.
If BinaryNLS were to parse the metadata each time it retrieves data from the buffer, updates would be correctly taken into account, right ?
Yet, I agree that given the external resource loader feature does not provide any synchronization mechanism, this is probably unsafe in most cases.
But for an external image in RAW format that keeps the same size (width & height) typically, I would think it should be possible to update the pixels colors in a relatively safe manner.
WDYT ?

apoudret-microej: i think that when an external resource is opened by an app, it should be be locked until the app closes the resource. like you would expect with a file (if the app has opened a file input stream, the file can't be modified). modifying the external resource while it is opened would depend on implementation specific behavior of the consumer

for an external image, i think it would not work in many cases (different size or different raw format) and when it works i don't think that the developer should rely on it

ApplicationDeveloperGuide/nls.rst
apoudret-microej: ok it makes sense. i think it is good now

ApplicationDeveloperGuide/nls.rst
rlouedec-microej: Ok, I am fine with making this is limitation of the external resources loader.

Even though the application consumer has full control on how to consume the resource, default open mode for modern file systems is "exclusive read" (https://stackoverflow.com/a/3448307).

Done: cfb1cbe842c21e2fe0a6f50bcce81fa84d06fa46

PR #676, opened 2023-08-17
Added a global list of build types per SDK version and added the list of latest buildtypes added per each SDK changelog entry

gjadi-mej:
I didn't fill the build types for SDK 5.0.1 because the tag Easyant4Eclipse v2.1.1 is missing.

### DIFF
--- a/SDKUserGuide/sdkChangeLog.rst
+++ b/SDKUserGuide/sdkChangeLog.rst
@@ -1038,6 +1038,208 @@ Skeletons
 -  Added the latest HIL Engine API to mock-up skeleton (Start
    and Stop listeners hooks)
 
+-------------------
+Build Types per SDK
+-------------------
+
+- SDK 5.8.0
+
+  - build-addon-processor#2.2.0
+  - build-application#9.2.0
+  - build-artifact-repository#3.4.0
+  - build-custom#2.2.0
+  - build-firmware-customizer#3.2.0
+  - build-firmware-multiapp#8.1.0
+  - build-firmware-singleapp#2.2.0
+  - build-izpack#3.3.0
+  - build-microej-extension#2.2.0
+  - build-microej-javaapi#5.2.0
+  - build-microej-javaimpl#5.2.0
+  - build-microej-javalib#6.2.0
+  - build-microej-mock#2.2.0
+  - build-microej-ri#3.2.0
+  - build-microej-testsuite#4.2.0
+  - build-product-java#2.2.0
+  - build-runtime-api#4.1.0
+  - build-std-javalib#3.2.0
+  - microej-meta-build#3.0.0
+- SDK 5.7.0
+
+  - build-addon-processor#2.2.0
+  - build-application#9.2.0
+  - build-artifact-repository#3.3.0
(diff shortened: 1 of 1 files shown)


SDKUserGuide/sdkChangeLog.rst
```diff
@@ -189,6 +189,8 @@ Build Types
 ^^^^^^^^^^^
 
 - Added Gradle dependency line in the generated Javadoc of an artifact repository (``build-artifact-repository`` build type).
```
tdelhomenie-microej, line 191: Shouldn't this line be under the line `Added build-artifact-repository#3.4.0` ?

SDKUserGuide/sdkChangeLog.rst
```diff
@@ -234,9 +236,27 @@ General
...
+- Added build-microej-javaimpl#5.2.0
+- Added build-microej-javalib#6.2.0 
+  - Fixed compilation with JDK 11.
+  - Fixed Artifact Checker's execution.
+- Added build-microej-mock#2.2.0
+- Added build-microej-ri#3.2.0
+- Added build-microej-testsuite#4.2.0
+- Added build-product-java#2.2.0
+- Added build-runtime-api#4.1.0
+- Added build-std-javalib#3.2.0
```
tdelhomenie-microej, line 259: It is weird to have new version of build types but without content, it seems like we did an upgrade for nothing. Shouldn't we instead have something like:
```
- Fixed ``build-std-javalib`` compilation with JDK 11.
- Fixed Artifact Checker's execution on ``build-std-javalib``.
- Fixed ``build-artifact-repository`` build type which couldn't find the previous release of the repository to merge it.
- Added new build types version:
  - build-addon-processor#2.2.0
  - build-application#9.2.0
  - ...
```
?

SDKUserGuide/sdkChangeLog.rst
```diff
@@ -1038,6 +1118,191 @@ Skeletons
 -  Added the latest HIL Engine API to mock-up skeleton (Start
    and Stop listeners hooks)
 
+------------------
+Buildtypes per SDK
+------------------
```
tdelhomenie-microej, line 1123: Shouldn't we put this section in a dedicated page instead?

tdelhomenie-microej (changes requested):

SDKUserGuide/sdkChangeLog.rst
gjadi-mej: As said on another channel, this is in the same source file but in a different html chapter in the final rendering (same as SDK Dist and SDK changelogs).

SDKUserGuide/sdkChangeLog.rst
gjadi-mej: I was also thinking about that, but to use the same format for all build type version declaration. To have a dedicated section `Latest Build Type Available`.

WDYT?


(Especially because it may not be true that the change mentioned was done in the version mentioned (e.g. it may have been done in an earlier version and all of them are added but we're only showing the latest)).

tdelhomenie-microej (approved):

PR #680, opened 2023-08-18
Use tabs to regroup common actions per env (CLI, IntelliJ, Eclipse)

gjadi-mej:
This is to showcase the use of tabs more widely than what is done currently. It the result is convincing, it should be done to other sections as well.

IMHO it greatly reduces the cognitive load when reading the documentation.


tdelhomenie-microej:
The result looks great 👍 It should be done everywhere.

gjadi-mej:
> The result looks great 👍 It should be done everywhere.

Added to all pages when it made sense.

SDK6UserGuide/install.rst
```diff
@@ -106,110 +106,106 @@ Install the IDE Plugin
...
+         :alt: IntelliJ IDEA Plugin Installation
+         :align: center
+         :scale: 70%
+      
+         IntelliJ IDEA Plugin Installation
+      
+      - Click on the ``Install`` button.
+      - Click on the ``Restart IDE`` button.
+      
+   .. tab:: IntelliJ IDEA Snapshot Version
```
tdelhomenie-microej, line 130: We should take this opportunity to move this section in another location. This is a special use case and we should keep the installation process as simple as possible. Maybe we can move it in a new how-to in https://docs.microej.com/en/feature-microej-sdk-6_test_tabs/SDK6UserGuide/howtos/index.html. WDYT?

SDK6UserGuide/howtos/installSnapshotPlugins.rst
```diff
@@ -0,0 +1,33 @@
+.. _sdk_6_install_plugin_snapshot:
+
+Install MicroEJ Plugin Snapshot Version on IntelliJ IDEA
```
tdelhomenie-microej, line 3:
```suggestion
How to Install MicroEJ Plugin Snapshot Version on IntelliJ IDEA
```

tdelhomenie-microej (changes requested):

tdelhomenie-microej (approved):

PR #681, opened 2023-08-21
Ignore unstable URLs in linkcheck

tdelhomenie-microej:


gjadi-mej:
Maybe add a comment to indicate why those links are being ignored. Otherwise it's inconsistent with the above comment.

tdelhomenie-microej:
Done.

gjadi-mej (approved):

PR #688, opened 2023-08-28
Use markeplace for the Eclipse plugin installation

tdelhomenie-microej:
https://docs.microej.com/en/feature-sdk6-eclipse-marketplace/SDK6UserGuide/install.html#install-the-ide-plugin


gjadi-mej (changes requested):
Please use
```
:guilabel:`label`
```
instead of code block for gui labels.

tdelhomenie-microej:
@gjadi-mej done

PR #696, opened 2023-08-31
Feature/m0081 meja 2050 fix debugger doc friviere review

frivieremicroej:


ApplicationDeveloperGuide/debuggerProxy.rst
```diff
@@ -89,19 +89,10 @@ Run the program:
...
-.. code-block:: sh
-
-    # generate core dump file
-    # the argument `file` is optional. It specifies the destination file where the core dump is saved
-    generate-core-file [file]
-
 .. warning::
-    This function must only be called from the MicroJvm virtual machine thread context and only from a native function or callback.
+    
+    The core dump should only be called when the Core Engine task is stopped on one of the specified callbacks or in a native function.
```
agazardMicroeJ, line 91: As we never mentioned the term callbacks in the previous lines, can we specified them in brackets like this ?

The core dump should only be called when the Core Engine task is stopped on one of the specified callbacks (`LLMJVM_on_OutOfMemoryError_thrown()`, `LLMJVM_on_Runtime_gc_done()` etc.) or in a native function.

PR #701, opened 2023-08-31
Feature/m0078 ddg 544 add gnss foundation documentation

tboue:
Adds documentation about GNSS Foundation Library.


frivieremicroej:
@tboue I see that the modules are published on forge.
I suggest to just roll-out on the Simulator and if OK merge it as it. The doc seems pretty clean. It's worth merging in the main stream for indicating it is available.

frivieremicroej (commented):
@tboue I see that the modules are published on forge.
I suggest the experience team to just roll-out on the Simulator and if OK merge it as it. The doc seems pretty clean. It's worth merging in the main stream for indicating it is available.

frivieremicroej (approved):

PR #722, opened 2023-09-15
Fix and improve SDK 6 doc about tests

tdelhomenie-microej:


SDK6UserGuide/testProject.rst
```diff
@@ -212,6 +212,11 @@ This can be convenient to quickly execute one test for example, without requirin
 Refer to the Gradle `Test filtering <https://docs.gradle.org/current/userguide/java_testing.html#test_filtering>`__
 documentation for more details on how to filter the tests and on the available patterns.
 
+.. warning::
+
+   At the moment, only class-level filtering is supported. 
+   This means that, for instance, you are unable to run a single test method within a test class.
```
gjadi-mej, line 218:
```suggestion
   This means that, for instance, it is not possible to run a single test method within a test class.
```

gjadi-mej (approved):
Approved, apply the suggestion if you're ok with it.

SDK6UserGuide/testProject.rst
tdelhomenie-microej: So you don't like what ChatGPT suggests? ;) It is because of the usage of "you"?

gjadi-mej: the `you` thing, yes. Not a formal rule, I know.

tdelhomenie-microej: Applied. Not sure too.

PR #752, opened 2023-10-02
Fix pointer manipulation issue + MISRA C:2012 compliancy

jcourcelmicroej:


VEEPortingGuide/architecture8Migration.rst
```diff
@@ -73,6 +73,7 @@ The code logic based on a ``malloc/free`` implementation does not need to be cha
 
    int32_t LLKERNEL_IMPL_getFeatureHandle(int32_t allocation_index) {
       // No persistency support
+      (void)allocation_index; // parameter not used by this implementation
```
frivieremicroej, line 76: do we really need that ?

PR #756, opened 2023-10-03
Improve import project page

tdelhomenie-microej:
https://docs.microej.com/en/feature-open-project/SDK6UserGuide/importProject.html


gjadi-mej (commented):
I think it would be nice to provide a screenshot for the Android setup too. (IntelliJ and Eclipse have screenshots).

In addition, could you align the 2 labels on IntelliJ screenshots?
> Fig 1. Project Model Selection when Opening in IntelliJ IDEA
> Fig 2. Project Model Selection when Importing in IntelliJ IDEA

Maybe by using images of the same size? Or another trick in rst...

tdelhomenie-microej:
@gjadi-mej I am not sure which screenshot we can add here (Android Studio never asks to select the project model). The only screenshot would be the "Browse" window, but we should add it also in IntelliJ and Eclipse and I am not sure it is relevant. WDYT?

> In addition, could you align the 2 labels on IntelliJ screenshots?

The only solution I found is to use raw HTML... quite ugly but it works. The drawbacks are that if the default figure style changes in the future the style of these images must be adapted, and the image is not clickable.

cblanchard-microej:
On my side, I do not see the pictures when you are using HTML. I got the following error: 

image

The path does not seems to work.

gjadi-mej:
> @gjadi-mej I am not sure which screenshot we can add here (Android Studio never asks to select the project model). The only screenshot would be the "Browse" window, but we should add it also in IntelliJ and Eclipse and I am not sure it is relevant. WDYT?

I think it would be better with the screenshot for IntelliJ and Android Studio. For Eclipse I think it's fine because there is already a screenshot when setting the root project dir.

> > In addition, could you align the 2 labels on IntelliJ screenshots?
> 
> The only solution I found is to use raw HTML... quite ugly but it works. The drawbacks are that if the default figure style changes in the future the style of these images must be adapted, and the image is not clickable.

For the MicroUI Flush Visualizer I used a table: https://github.com/MicroEJ/docs/blob/master/ApplicationDeveloperGuide/UI/Advanced/how-to-microui-flush-visualizer.rst?plain=1#L18
#oldSchoolHTML

tdelhomenie-microej:
> I think it would be better with the screenshot for IntelliJ and Android Studio. For Eclipse I think it's fine because there is already a screenshot when setting the root project dir.

Done.

> For the MicroUI Flush Visualizer I used a table: https://github.com/MicroEJ/docs/blob/master/ApplicationDeveloperGuide/UI/Advanced/how-to-microui-flush-visualizer.rst?plain=1#L18
#oldSchoolHTML

That's what I initially used (list-table) but that did not allow to format the image caption correctly. I can (almost) do what I want with plain old HTML.

@cblanchard-microej it should be better now, let me know.

cblanchard-microej:
Yes it is working on my side. 
As @gjadi-mej said, maybe you could add the "Select root directory" screenshot for IntelliJ?

tdelhomenie-microej:
> As @gjadi-mej said, maybe you could add the "Select root directory" screenshot for IntelliJ?

@cblanchard-microej done

cblanchard-microej (approved):
It is good for me

gjadi-mej (approved):

PR #773, opened 2023-10-12
Document gradle auto-reloading

denisbedin:
https://docs.microej.com/en/feature-microej-sdk-6-gradle-autoreloading/SDK6UserGuide/howtos/gradleAutoreloading.html


SDK6UserGuide/howtos/gradleAutoreloading.rst
```diff
@@ -0,0 +1,75 @@
...
+      - Click on :guilabel:`File` > :guilabel:`Settings...`.
+      - Go to  :guilabel:`Build, Execution, Deployment` > :guilabel:`Build Tools`.
+      - The :guilabel:`Reload changes in the build scripts` option is enabled by default. Check the :guilabel:`Any changes` option.
+
+      .. figure:: images/intellij-autoreload.png
+         :alt: auto-reload option in IntelliJ IDEA
+         :align: center
+         :scale: 70%
+      
+         Auto-reload option in IntelliJ IDEA
```
bguedas, line 51: Why don't we also check reload options for Kotlin scripts in IDEA as in Android Studio?

bguedas (changes requested):
Don't forget to wait that the branch you derived from has been merged before merging, or select this branch as target

SDK6UserGuide/howtos/gradleAutoreloading.rst
```diff
@@ -0,0 +1,75 @@
+.. _sdk_6_howto_gradle_autoreloading:
+
+How To Automatically reload a Gradle project
+============================================
+
+By default, regardless of the IDE that you are using (Android Studio, IntelliJ IDEA or Eclipse), any file change is automatically saved but
```
tdelhomenie-microej, line 6: Are you sure it is the case for Eclipse?

SDK6UserGuide/howtos/gradleAutoreloading.rst
```diff
@@ -0,0 +1,75 @@
+.. _sdk_6_howto_gradle_autoreloading:
+
+How To Automatically reload a Gradle project
+============================================
+
+By default, regardless of the IDE that you are using (Android Studio, IntelliJ IDEA or Eclipse), any file change is automatically saved but 
+the reload of a Gradle project must be explicitly triggered by the user when the configuration of the project has changed.
```
tdelhomenie-microej, line 7: We could add a sentence explaining that in this case the advantage is that it avoids to reload the project too frequently (at each change), so it should be less resources consuming, but the drawback is that the developer must not forget to reload the project manually to apply the changes.
It can help the user decide which is the best option for their case.

SDK6UserGuide/howtos/gradleAutoreloading.rst
denisbedin: I didn't need to enable them to make the auto-reloading work. But It's better to check them, so I merged the Android Studio and IntelliJ IDEA tabs.

SDK6UserGuide/howtos/gradleAutoreloading.rst
denisbedin: Done.

SDK6UserGuide/howtos/gradleAutoreloading.rst
denisbedin: I modified the sentence to remove the "file change is automatically saved" part.

tdelhomenie-microej (approved):

bguedas (approved):

PR #784, opened 2023-10-20
Fix project creation with Android Studio

tdelhomenie-microej:


bguedas (approved):

thomasponsmicroej:
Feels right to me, maybe it is a dumb idea but if we want to share some doc between the sdk user guide and the vee wear ack guide, do you think an altenrative path could be:
1 create a project with empty activity (like now)
2 delete subproject app
3 file > new module
4 steps similar to intellij way

We could point to steps 3 and 4 form our doc our use it as include.
wdyt?

tdelhomenie-microej:
@thomasponsmicroej as discussed, the idea is interesting but it does not simplify the creation process (app module must be remove in the Module Settings windows, then deleted, then a lib module must be created and completely changed again), so we keep the first version.

PR #791, opened 2023-10-25
Feature/m0090 ide 4443 doc stack trace reader sdk6

gjadi-mej:

PR #793, opened 2023-10-26
Feature/m0090 ide 4446 Add code coverage analyzer and how to enable code coverage cc generation

gjadi-mej:
Based on #791

### DIFF
--- a/SDK6UserGuide/codeCoverageAnalyzer.rst
+++ b/SDK6UserGuide/codeCoverageAnalyzer.rst
@@ -1,0 +1,133 @@
+.. _sdk6.section.code_coverage_analyzer:
+
+======================
+Code Coverage Analyzer
+======================
+
+
+Principle
+=========
+
+The Simulator features an option to output ``.cc`` (Code Coverage)
+files that represent the use rate of functions of an application. It
+traces how the opcodes are really executed.
+
+Functional Description
+======================
+
+The Code Coverage Analyzer scans the output ``.cc`` files, and outputs an
+HTML report to ease the analysis of methods coverage. The HTML report is
+available in a folder named htmlReport in the same folder as the .cc
+files.
+
+Dependencies
+============
+
+In order to work properly, the Code Coverage Analyzer should input the ``.cc`` files (See :ref:`sdk_6_generate_code_coverage`).
+The ``.cc`` files rely on the classpath used during the execution of the Simulator to the Code Coverage Analyzer.
+Therefore the classpath is considered to be a dependency of the Code Coverage Analyzer.
+
+Installation
+============
+
+This tool is a built-in Architecture tool.
+
+Use
+===
+
+A MicroEJ tool is available to launch the Code Coverage Analyzer tool.
+The tool name is Code Coverage Analyzer.
+
+Two levels of code analysis are provided, the Java level and the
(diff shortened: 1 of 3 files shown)


SDK6UserGuide/codeCoverageAnalyzer.rst
```diff
@@ -0,0 +1,115 @@
...
+       --toolProperty="cc.includes=" \
+       --toolProperty="cc.excludes=" \
+       --toolProperty="cc.src.folders=/MODULE_PATH/src" \
+       --toolProperty="cc.html.dir=/MODULE_PATH/cc"
+
+Options
+=======
+
+Option: \*.cc files folder
+--------------------------
```
bguedas, line 59: This Option section is not of the same kind as the others: `------` instead of `^^^^^^`

SDK6UserGuide/codeCoverageAnalyzer.rst
```diff
@@ -0,0 +1,115 @@
...
+::
+
+   gradle :execTool --name=codeCoverageAnalyzer \
+       --toolProperty="cc.dir=/MODULE_PATH/build/output/com.microej.Main/cc/" \
+       --toolProperty="cc.includes=" \
+       --toolProperty="cc.excludes=" \
+       --toolProperty="cc.src.folders=/MODULE_PATH/src" \
+       --toolProperty="cc.html.dir=/MODULE_PATH/cc"
+
+Options
```
bguedas, line 55: Are they really options or arguments? I mean, an "option" is supposed to be optional. I guess that some of them are mandatory and other optional. Moreover, we do not describe `cc.src.folders` and `cc.html.dir`.

SDK6UserGuide/runOnSimulator.rst
```diff
@@ -108,9 +108,34 @@ If you want to connect the IDE debugger:
...
+
+*Option Name*: ``s3.cc.activated``
+
+*Description*
+
+Set to ``true`` to enable the generation of Code Coverage files, don't define the property to disable the generation.
+
 ..
-   | Copyright 2008-2023, MicroEJ Corp. Content in this space is free 
+   | Description 2008-2023, MicroEJ Corp. Content in this space is free
```
bguedas, line 138: Really?

bguedas (changes requested):

SDK6UserGuide/runOnSimulator.rst
gjadi-mej: Fixed. No idea how I did this :D

SDK6UserGuide/codeCoverageAnalyzer.rst
gjadi-mej: Fixed.

SDK6UserGuide/codeCoverageAnalyzer.rst
gjadi-mej: I've added description for src.folders and html.dir.

I kept the name options because that's what is used in the SDK5 documentation.
Also, for now they are required but as soon as we will provide dedicated task, we will be able to provide the suitable default value (as it's done in SDK5).

Not sure it's worth changing the formulation.

bguedas (approved):

PR #806, opened 2023-11-20
M0090IDE-4475 Add documentation for the Eclipse plugin tools

tdelhomenie-microej:
https://docs.microej.com/en/feature-m0090ide-4475_eclipse-tools/


SDK6UserGuide/memoryMapAnalyzer.rst
```diff
@@ -0,0 +1,67 @@
...
+
+===================
+Memory Map Analyzer
+===================
+
+Principle
+=========
+
+When the Executable of an Application is built, a Memory Map file is generated. 
+This file can be visualized with the The Memory Map Analyzer, an Eclipse IDE plug-in.
```
denisbedin, line 11:
```suggestion
This file can be visualized with the Memory Map Analyzer, an Eclipse IDE plug-in. 
```

SDK6UserGuide/tools.rst
```diff
@@ -17,14 +21,46 @@ The optional parameters ``--toolProperty`` are used to configure the tool's opti
...
+         - In the upcoming window, click on the :guilabel:`Restart Now` button.
+
+The following sections describe the IDE tools and their options:
+
+.. toctree::
+   :maxdepth: 1
+
+   memoryMapAnalyzer
+   heapAnalyzer
+   ../ApplicationDeveloperGuide/UI/MicroUI/fontDesigner
```
denisbedin, line 62: The Font Designer page should be in ` ../ApplicationDeveloperGuide/UI/MicroUI/fontDesigner` but it appears in the SDK 6 User Guide section : 
Capture

Is this normal ?

tdelhomenie-microej: It is probably because there is no entry in the UI ToC for this tool, so it displays the one from the SDK  section. I don't think it is a problem actually. WDYT?

SDK6UserGuide/memoryMapAnalyzer.rst
tdelhomenie-microej: Done.

SDK6UserGuide/tools.rst
denisbedin: I didn't notice that there is no entry in the UI ToC, so it seemed weird, but it's ok.

PR #820, opened 2023-12-04
clarify documentation about Composite builds

denisbedin:
https://docs.microej.com/en/feature-sdk6_clarify_composite_build/SDK6UserGuide/howtos/resolveInIDE.html

### DIFF
--- a/SDK6UserGuide/howtos/resolveInIDE.rst
+++ b/SDK6UserGuide/howtos/resolveInIDE.rst
@@ -41,7 +41,49 @@ This can be configured by adding the following line in the ``settings.gradle.kts
 
 The path given to the ``includeBuild`` method is the relative path of the project to include.
 
+.. warning::
+   The ``includeBuild`` method should be used to declare a dependency between two autonomous projects.
+   To declare a dependency between two subprojects of a multi-project, a `Project dependency <https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:project_dependencies>`__ must be used.
+   Refer to the :ref:`sdk6_multiproject_dependencies` section for more information.
+
 Refer to the `Official Gradle documentation on the Composite Build feature <https://docs.gradle.org/current/userguide/composite_builds.html>`__ for more details.
+
+.. _sdk6_multiproject_dependencies:
+
+Dependencies Between Subprojects of a Multi-Project
+---------------------------------------------------
+
+Gradle allows to declare dependencies between subprojects of a `multi-project build <https://docs.gradle.org/current/userguide/multi_project_builds.html>`__ 
+by declaring a `Project dependency <https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:project_dependencies>`__.
+
+For example, if you have a multi-project named ``myProject`` composed of two subprojects ``myApplication`` and ``myLibrary``:
+
+.. code-block::
+
+   |- myProject
+   |   |- myApplication
+   |   |   |- src
+   |   |   |- build.gradle.kts
+   |   |   |- settings.gradle.kts
+   |   |- myLibrary
+   |   |   |- src
+   |   |   |- build.gradle.kts
+   |   |   |- settings.gradle.kts
+   |   |- settings.gradle.kts
+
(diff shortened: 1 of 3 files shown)


SDK6UserGuide/howtos/resolveInIDE.rst
```diff
@@ -41,8 +41,48 @@ This can be configured by adding the following line in the ``settings.gradle.kts
 
 The path given to the ``includeBuild`` method is the relative path of the project to include.
 
+.. warning::
+   The ``includeBuild`` method must be used to declare a dependency between two autonomous projects.
+   To declare a dependency between two subjects of a multi-project, a `Project dependency <https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:project_dependencies>`__ must be used.
```
bguedas, line 46:
```suggestion
   To declare a dependency between two subprojects of a multi-project, a `Project dependency <https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:project_dependencies>`__ must be used.
```

bguedas, line 46: I don't think that "**must**" should be used in the two lines above because it is not mandatory. We should use "**should**" instead because it is an advice.

SDK6UserGuide/howtos/resolveInIDE.rst
```diff
@@ -41,8 +41,48 @@ This can be configured by adding the following line in the ``settings.gradle.kts
...
+   Refer to the :ref:`sdk6_multiproject_dependencies` section for more information.
+
 Refer to the `Official Gradle documentation on the Composite Build feature <https://docs.gradle.org/current/userguide/composite_builds.html>`__ for more details.
 
+.. _sdk6_multiproject_dependencies:
+
+Dependencies between subprojects of a multi-project
```
bguedas, line 53:
```suggestion
Dependencies Between Subprojects of a Multi-Project
```

SDK6UserGuide/howtos/resolveInIDE.rst
```diff
@@ -41,8 +41,48 @@ This can be configured by adding the following line in the ``settings.gradle.kts
...
+
+You can declare a Project dependency in the ``build.gradle.kts`` file of the ``myApplication`` subproject to 
+make it depend on the ``myLibrary`` subproject::
+
+   dependencies {
+      implementation(project(":myLibrary"))
+   }
+
+This allows to avoid having to build and publish the ``myLibrary`` subproject then refresh dependencies on 
+the ``myApplication`` project to get the update whenever you do a change in the ``myLibrary`` project.
```
bguedas, line 82: This sentence is too long and I'm not sure to get its meaning. The problem may only the use of "then" and the lack of punctuation. Maybe something like "This allows to avoid having to build and publish the ``myLibrary`` subproject, and then to refresh dependencies on the ``myApplication`` project to get the update, whenever you do a change in the ``myLibrary`` project.". I would also say "everytime" instead of "whenever".

bguedas (changes requested):

SDK6UserGuide/howtos/resolveInIDE.rst
denisbedin: Done.

denisbedin: Done.

SDK6UserGuide/howtos/resolveInIDE.rst
denisbedin: Done.

SDK6UserGuide/howtos/resolveInIDE.rst
denisbedin: I modified the sentence, is it better now ?

denisbedin:
@bguedas I also had to change the Eclipse link that we use in the SDK 6 User Guide, https://www.eclipse.org/downloads/packages/release/2022-09/r/eclipse-ide-java-developers is broken. I replaced it by https://www.eclipse.org/downloads/packages, I think it's better to reference the latest Eclipse release instead of the minimum version that we support.

PR #837, opened 2023-12-22
document project creation with Android Studio wizards

denisbedin:
https://docs.microej.com/en/feature-sdk6_android_studio_project_wizards/


SDK6UserGuide/install.rst
```diff
@@ -125,36 +125,66 @@ Once your favorite IDE is installed, the MicroEJ plugin must be installed.
...
+            :alt: Android Studio Plugin Installation - Third-Party Plugins Notice
+            :align: center
+            :scale: 70%
+         
+            Android Studio Plugin Installation - Third-Party Plugins Notice
+
+      - Click on the :guilabel:`Restart IDE` button.
+
+      To install the snapshot version of the MicroEJ plugin, please refer to :ref:`sdk_6_install_plugin_snapshot`.
+
```
bguedas, line 157: I think that we can add a warning at the end of this tab to tell that the MicroEJ plugin should be uninstalled if it has been previously installed. Something like

      .. warning::

         There used to be a unique plugin for both Android Studio and IntelliJ IDEA. There are now dedicated plugins.
         If ``MicroEJ`` plugin has been previously installed, you should uninstall it and install ``MicroEJ for Android Studio`` instead.

bguedas (changes requested):

SDK6UserGuide/install.rst
denisbedin: Done.

bguedas (approved):

PR #848, opened 2024-01-09
execTool: Document Local Deployment Socket

gjadi-mej:


SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+=========
+
+The Local Deployment Socket is a tool that allows to transfer a Sandboxed Application on the device over a network connection.
+
+Functional Description
+======================
+
+The Local Deployment Socket builds the Sandboxed Application ``.fo`` and upload it on the device identified by its IP address. On the device, it is the job of the Kernel Application that receives the ``.fo`` to install and to run the Sandboxed Application.
+
+Dependencies
```
bguedas, line 17: It looks like a section from a README. Is it the case? And is it relevant here, especially when there is no dependency?

SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+============
+
+No dependency.
+
+Installation
+============
+
+This tool is a built-in Architecture tool bundled in the VDE when it is built.
```
bguedas, line 25: I can't find any other VDE occurence in the doc, I thoink we should say Virtual Device instead.

SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+Option: Application Feature Class
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+*Option Name*: ``application.main.class``
+
+*Required?*: Yes
+
+*Description*:
+
+Specify the FQDN of the Feature Application to deploy. It is the entry-point of the feature.
```
bguedas, line 51: Maybe explain FQDN in full words

bguedas (changes requested):

SDK6UserGuide/localDeployTool.rst
gjadi-mej: Not from a README but I reused a similar format as the stack trace reader.
No strong opinion on this.

SDK6UserGuide/localDeployTool.rst
gjadi-mej: updated.

SDK6UserGuide/localDeployTool.rst
gjadi-mej: updated.

SDK6UserGuide/localDeployTool.rst
bguedas: Neither do I, I think the best thing is to be consistent for each tool. Let's see what the others say

tdelhomenie-microej: I don't see the value to keep it, so I would remove it.

SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Principle
+=========
+
+The Local Deployment Socket is a tool that allows to transfer a Sandboxed Application on the device over a network connection.
+
+Functional Description
+======================
+
+The Local Deployment Socket builds the Sandboxed Application ``.fo`` and upload it on the device identified by its IP address. On the device, it is the job of the Kernel Application that receives the ``.fo`` to install and to run the Sandboxed Application.
```
tdelhomenie-microej, line 15: Technically, the tool does not build the `.fo` file, it only deploys it. Not a big issue, but we could replace this by something like "The Local Deployment Socket makes sure that the Sandboxed Application ``.fo`` is built and upload it ..." instead?

SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Dependencies
+============
+
+No dependency.
+
+Installation
+============
+
+This tool is a built-in Architecture tool bundled in the Virtual Device when it is built.
```
tdelhomenie-microej, line 25: This is an implementation detail IMO, and it is not totally true since the tool is also composed of the execTool Gradle task, which is part of the SDK. Shouldn't we say instead that this tool comes with the SDK, thanks to the `execTool` Gradle task?

SDK6UserGuide/tools.rst
```diff
@@ -60,6 +60,7 @@ The following sections describe the IDE tools and their options:
    memoryMapAnalyzer
    heapAnalyzer
    ../ApplicationDeveloperGuide/UI/MicroUI/fontDesigner
+   localDeployTool
```
tdelhomenie-microej, line 63: I know it is designed as a MicroEJ Tool, but it seems strange to me that deploying a Standalone Application is done with a runOnDevice task, but not a Sandboxed Application. This brings confusion IMO. That's not the purpose here, but we should probably think of allowing to deploy a fo file with the runOnDevice task.

SDK6UserGuide/localDeployTool.rst
```diff
@@ -0,0 +1,103 @@
...
+
+*Required?*: Yes
+
+*Description*:
+
+A boolean describing whether to use the storage to store the Application or not.
+Refer to the Kernel documentation to find out the correct setting.
+
+..
+   | Copyright 2024, MicroEJ Corp. Content in this space is free
```
tdelhomenie-microej, line 99: I think all the documentation pages should all have the same years, so "2008-2024".

SDK6UserGuide/tools.rst
```diff
@@ -60,6 +60,7 @@ The following sections describe the IDE tools and their options:
...
 
 ..
    | Copyright 2023, MicroEJ Corp. Content in this space is free
```
tdelhomenie-microej, line 66: Wrong year.

tdelhomenie-microej (changes requested):

SDK6UserGuide/localDeployTool.rst
gjadi-mej: As far as I can see, yes the tool does the build of the .fo. There is no dependency from the `:execTool` task to the `:buildFeature` task. It's all done under the hood by `localDeploymentSocket.microejLaunch`.

tdelhomenie-microej: Waouh ok, I did not know! So there are 2 ways to build a fo file... Are we sure it is built with the same options in both cases?

gjadi-mej: Hmm I didn't think about that. I'll have to check it out.

SDK6UserGuide/tools.rst
gjadi-mej: agreed. I think it could be done once we've documented the basics of the microejtool and we will work on improve the experience with dedicated tasks (or here, with well integrated task).

SDK6UserGuide/localDeployTool.rst
gjadi-mej: I don't know. Maybe we could just remove the section because it doesn't bring anything. I'm not sure.
I've reused the same structure as the stacktrace reader

For the stack trace reader we have:

> This tool is a built-in Architecture tool.

We can of course update the stack trace reader doc too.

tdelhomenie-microej: I agree, we should remove them in both cases.

SDK6UserGuide/tools.rst
tdelhomenie-microej: Agreed.

SDK6UserGuide/localDeployTool.rst
gjadi-mej: fixed.

SDK6UserGuide/localDeployTool.rst
gjadi-mej: fixed.

SDK6UserGuide/localDeployTool.rst
gjadi-mej: I've aligned the behavior of the buildFeature task with that of the localDeploymentSocket tool in the gradle plugin (in review).

tdelhomenie-microej (approved):

bguedas (approved):

PR #855, opened 2024-01-19
Move Heap Usage monitoring doc page to Application Dev Guide

gjadi-mej:
https://docs.microej.com/en/feature-m0090ide-4563/ApplicationDeveloperGuide/heapUsageMonitoring.html


ApplicationDeveloperGuide/heapUsageMonitoring.rst
```diff
@@ -141,7 +141,7 @@ To analyze heap usage and see what objects are alive in the application, use the
 :ref:`heapdumper` tools.
```
tdelhomenie-microej, line 141: This link is SDK 5. We should add both SDK 5 and SDK 6 links.

tdelhomenie-microej (changes requested):

ApplicationDeveloperGuide/heapUsageMonitoring.rst
gjadi-mej: Good catch, updated.

PR #877, opened 2024-02-12
Resolve local MMM modules from SDK 6

rlouedec-microej:
This should help on the migration to SDK 6 (mix SDK 5 / SDK 6 environment).

### DIFF
--- a/SDK6UserGuide/migrateMMMProject.rst
+++ b/SDK6UserGuide/migrateMMMProject.rst
@@ -181,7 +181,7 @@ will be converted to::
 
    testImplementation("ej.library.test:junit:1.7.1")
 
-Also note that this will not resolve snapshot builds. To resolve both snapshot and release versions, use ``[1.0.0-rc,1.0.0]`` instead of ``1.0.0``.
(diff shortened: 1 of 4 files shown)


rlouedec-microej:
I know there is already a note about snapshot versions in https://docs.microej.com/en/latest/SDK6UserGuide/addDependency.html#version
but I think it is good to mention in the migration guide as well.

rlouedec-microej:
Also, what about moving the "Migrate an MMM project" to the "Migration Notes" section ? (I think that is not the first time I went there when looking for that page)

SDK6UserGuide/migrateMMMProject.rst
```diff
@@ -181,6 +181,12 @@ will be converted to::
 
    testImplementation("ej.library.test:junit:1.7.1")
 
+Also note that this will not resolve snapshot builds. To resolve both snapshot and release versions, use ``[1.0.0-rc,1.0.0]`` instead of ``1.0.0``.
```
tdelhomenie-microej, line 184:
```suggestion
Also note that this will not resolve snapshot builds since versions are explicit in SDK 6, see this chapter for more details. To resolve both snapshot and release versions, use ``[1.0.0-RC,1.0.0]`` instead of ``1.0.0``.
```
"this chapter" should link to https://docs.microej.com/en/latest/SDK6UserGuide/addDependency.html#version

tdelhomenie-microej (approved):

PR #882, opened 2024-02-16
update the CCO nema options

gbalan-microej:


seonmej (approved):

seonmej (changes requested):
`Parallel` in both diagrams should be replaced by `Single`.

PR #913, opened 2024-03-11
document architecture characteristics

frivieremicroej:
they are available as constants
Also inline the typical code for printing formatted charcteristics on standard output. This is the code embedded in the default pre-built app.

### DIFF
--- a/ApplicationDeveloperGuide/runtime.rst
+++ b/ApplicationDeveloperGuide/runtime.rst
@@ -300,10 +300,6 @@ The following code prints the formatted Architecture characteristics on standard
       System.out.println("- Instruction Set Architecture: " + isaStr);
       System.out.println("- Compilation Toolchain:        " + toolchainName + " (" + toolchainFullName + ")");
 
-.. note::
-
-   As of :ref:`Architecture 8.1.0 <changelog-8.1.0>`, a default pre-built application embeds the previous code.
-
 ..
    | Copyright 2008-2024, MicroEJ Corp. Content in this space is free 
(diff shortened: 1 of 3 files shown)


ApplicationDeveloperGuide/runtime.rst
```diff
@@ -243,6 +243,66 @@ On Architecture ``7.x`` you will get the following default error message:
...
+      System.out.println("- Name:                         " + name);
+      System.out.println("- Version:                      " + version + " (" + buildLabel + ")");
+      System.out.println("- Usage:                        " + usageStr);
+      System.out.println("- Core Engine Capability:       " + capabilityStr + "-Sandbox");
+      System.out.println("- Instruction Set Architecture: " + isaStr);
+      System.out.println("- Compilation Toolchain:        " + toolchainName + " (" + toolchainFullName + ")");
+
+.. note::
+
+   As of :ref:`Architecture 8.1.0 <changelog-8.1.0>`, a default pre-built application embeds the previous code.
```
bguedas, line 305: I'm not sure this note is really relevant because it's quite confusing. The code is embedded since `8.1.0` but I don't see how it is useful for the user who wants do know how to get this information.

bguedas (approved):

ApplicationDeveloperGuide/runtime.rst
frivieremicroej: Agree, I added a dedicated chapter for default application that links to Architecture characteristics

bguedas (approved):

PR #991, opened 2024-07-03
document evaluation license tool

frivieremicroej:


SDK6UserGuide/licenses.rst
```diff
@@ -136,6 +136,55 @@ The license key zip file must be simply dropped to the ``~/.microej/licenses/``
...
+
+To get more details on installed Evaluation licenses, run the debug tool as following:
+
+#. Open a terminal.
+#. Change directory to an Evaluation VEE Port.
+#. Execute the command:
+   
+   .. code:: console
+
+      java -Djava.library.path=resources/os/[OS_NAME] -jar licenseManager/licenseManagerKeyHardware.jar
```
denisbedin, line 154: When I run this command, it fails because the Main class is not found. It is not set in the MANIFEST.MF of the `licenseManagerKeyHardware.jar`. Will it be available in the next Architecture release ? If so, we should add a note to indicate that it requires a minimum Architecture version.

frivieremicroej: done

denisbedin (approved):

PR #1006, opened 2024-07-29
Explain how to specify the managed C function name if it is different…

agazardMicroeJ:
… from the java native method name

### DIFF
--- a/ApplicationDeveloperGuide/managedc/getting_started.rst
+++ b/ApplicationDeveloperGuide/managedc/getting_started.rst
@@ -62,6 +62,7 @@ To use Managed C in your Application, follow these steps:
             
             @Target(ElementType.METHOD)
             public @interface ManagedCFunction {
+               String value() default "";
             }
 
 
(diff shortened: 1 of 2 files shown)


ApplicationDeveloperGuide/managedc/getting_started.rst
```diff
@@ -125,7 +126,35 @@ To use Managed C in your Application, follow these steps:
...
+
+      The Managed C function name can be specified in the ``@ManagedCFunction`` annotation.
+      
+      .. code-block:: java
+         :emphasize-lines: 6
+
+         @ManagedCModule("factorial.wasm")
+         public class Main {
+
+             ...
```
dalgassimou, line 141: can you add an example of Java code which call the java native, Just to illustrate that the Java side use the Java name not the C name

ApplicationDeveloperGuide/managedc/getting_started.rst
```diff
@@ -125,7 +126,37 @@ To use Managed C in your Application, follow these steps:
             }
         }
 
-   .. note:: The Managed C function name and signature in the C file match the Java annotated native method.
+   .. note::
```
frivieremicroej, line 129: This shall not be part of the getting started, this is part of the specification
should be moved to `Calling Managed C from Java` section.

frivieremicroej (changes requested):

PR #1033, opened 2024-09-13
add  to CPU control specification the behaviour

frivieremicroej:
when an app is paused in a monitor

### DIFF
--- a/KernelDeveloperGuide/kf.rst
+++ b/KernelDeveloperGuide/kf.rst
@@ -707,7 +707,7 @@ A Kernel can assign an execution quota to a Feature using
 `Feature.setExecutionQuota()`_. The quota is expressed in execution
 units.
 
-Quotas account to the running current context owner.
+Quotas account to the context of the current thread's owner.
 
 When a Feature has reached its execution quota, its execution is
 suspended until all other Features have reached their execution quota.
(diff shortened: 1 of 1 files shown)


KernelDeveloperGuide/kf.rst
```diff
@@ -717,6 +717,11 @@ the execution counter of all Features is reset.
 Setting a Feature execution quota to zero causes the Feature to be
 suspended (the Feature is paused).
 
+If a Feature has reached its execution quota and is holding a monitor (through one of its threads), 
+and another Module (Feature or Kernel) with a disabled quota attempts (through one of its threads) to acquire the same monitor,
```
jerome-leroux, line 721:
```suggestion
and another Module (Feature or Kernel) with a disabled quota attempts to acquire the same monitor (through one of its threads), 
```

jerome-leroux (changes requested):
Line "Quotas account to the running current context owner." needs to be reviewed.

A correct statement would be "Quotas account to the context of the current thread's owner."

KernelDeveloperGuide/kf.rst
jerome-leroux, line 1: Line "Quotas account to the running current context owner." needs to be reviewed.

A correct statement would be "Quotas account to the context of the current thread's owner."

jerome-leroux (approved):

PR #1044, opened 2024-09-24
M0127CC-770 SDK 5 and 6 clarification for the overview

gharelMicroEJ:


acolleux-microej:
I would recommend to use the `ref` attribute to refer to other documentation pages instead of HTML pages.
For example use **:ref:`chapter-glossary`** instead of **`Glossary <glossary.html>`_**.
That way we make sure that the titles remain up to date.

acolleux-microej (approved):

PR #1055, opened 2024-10-02
Add example for SNI String conversions

rlouedec-microej:
https://microej-docs--1055.com.readthedocs.build/en/1055/VEEPortingGuide/sni.html#strings


gjadi-mej:
It would be interesting to also explain how to share the same value between the App and the VEE Port. Here the constant  MAX_STR_SIZE is duplicated.

gjadi-mej (approved):

PR #1057, opened 2024-10-03
Upgrade Gradle distribution from 8.0.2 to 8.6.

bguedas:


SDK6UserGuide/createProject.rst
```diff
@@ -543,9 +543,9 @@ It is also possible to add the Wrapper to an existing project by executing the `
   gradle wrapper
 
 The Gradle version used by the project can then be updated in the ``gradle/wrapper/gradle-wrapper.properties`` file. 
-The SDK requires Gradle ``8.0.2`` or higher::
+The SDK requires Gradle ``8.6`` or higher::
 
-  distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-bin.zip
+  distributionUrl=https\://services.gradle.org/distributions/gradle-8.6-bin.zip
```
denisbedin, line 548: In the tabs that describe the project creation, each IDE has a note about the version used by the Gradle wrapper. These notes must be updated:
- Eclipse and IntelliJ IDEA use Gradle 8.10.2
- Android Studio Koala uses Gradle 8.7 but I don't know for the other distributions, so maybe the note must be updated to indicate that the user must check if the wrapper uses at least 8.6.

denisbedin (changes requested):

SDK6UserGuide/createProject.rst
bguedas: I've updated the tabs

SDK6UserGuide/createProject.rst
```diff
@@ -69,7 +69,8 @@ This chapter explains the different ways to create a new project.
          Project View in Android Studio
 
       .. note::
-         The newly created Gradle project uses Gradle Wrapper with Gradle version ``8.2``.
+         The newly created Gradle project with Android Studio Koala (``2024.1.2``) uses Gradle Wrapper with Gradle version ``8.7``.
```
denisbedin, line 72: I don't think it is necessary to talk about Android Studio Koala, I think something like "Make sure that the newly created project uses at least Gradle Wrapper version ``8.6``" is enough.

denisbedin (changes requested):

SDK6UserGuide/createProject.rst
bguedas: This part has been rephrased.

denisbedin (approved):

PR #1058, opened 2024-10-03
Feature/audio

apoudret-microej:
https://docs.microej.com/en/feature-audio/

### DIFF
--- a/ApplicationDeveloperGuide/audio.rst
+++ b/ApplicationDeveloperGuide/audio.rst
@@ -1,0 +1,187 @@
+.. _audio_api:
+
+Audio
+=====
+
+Introduction
+------------
+
+The Audio API Library provides APIs to record and play audio in an Application.
+
+This API can be used in various use-cases, such as:
+
+- playing a sound
+- playing a music streamed over IP or Bluetooth
+- making a call
+- synthesizing speech (text-to-speech)
+- recognizing speech (speech-to-text)
+- using a voice assistant such as Alexa or ChatGPT
+
+.. _audio_api_usage:
+
+Usage
+-----
+
+The Audio API Library is provided as a Foundation Library.
+
+To use the `Audio API Library <https://repository.microej.com/modules/ej/api/audio/>`_, add the following line to the project build file:
+
+.. tabs::
+
+   .. tab:: Gradle (build.gradle.kts)
+
+      .. code-block:: kotlin
+
+         implementation("ej.api:audio:1.0.0")
+
+   .. tab:: MMM (module.ivy)
+
+      .. code-block:: xml
+
+         <dependency org="ej.api" name="audio" rev="1.0.0"/>
+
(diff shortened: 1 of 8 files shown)


thomasponsmicroej (approved):

ApplicationDeveloperGuide/audio.rst
```diff
@@ -0,0 +1,187 @@
...
+For example, the following snippet records audio with an audio record and writes the audio data to a file:
+
+.. code-block:: java
+
+	try (OutputStream outputStream = new FileOutputStream("record.raw")) {
+	    try (AudioRecord audioRecord = new AudioRecord(FORMAT, 1600)) {
+	        byte[] buffer = new byte[480];
+	        while (true) {
+	            audioRecord.readBuffer(buffer, 0, buffer.length); // read from audio record
+	            outputStream.write(buffer, 0, buffer.length); // write to file
```
gjadi-mej, line 90: Write the number of read data only?
```suggestion
	            int bytesRead = audioRecord.readBuffer(buffer, 0, buffer.length); // read from audio record
	            outputStream.write(buffer, 0, bytesRead); // write to file
```

VEEPortingGuide/bluetooth.rst
```diff
@@ -47,7 +39,7 @@ the following dependency to the :ref:`module.ivy <mmm_module_description>` file:
 
 ::
 
-	<dependency org="com.microej.pack.bluetooth" name="bluetooth-pack" rev="2.2.1" />
+	<dependency org="com.microej.pack.bluetooth" name="bluetooth-pack" rev="2.4.1" />
 
 The Platform project must be rebuilt (:ref:`platform_build`).
```
gjadi-mej, line 44: While you're there, you can remove this line.

gjadi-mej (changes requested):
Nice description of the API.

I've never been a fan of the description of the LLAPI as we don't really document them besides the header files...

ApplicationDeveloperGuide/audio.rst
apoudret-microej: readBuffer() blocks until the buffer has been fully filled. the number of bytes returned will be different from buffer size only if the audio record is closed in another thread during the read. in this sample it does not happen, so i think it is not necessary?

VEEPortingGuide/bluetooth.rst
apoudret-microej: fixed (amend)

ApplicationDeveloperGuide/audio.rst
gjadi-mej: I would say it's better to use the safer pattern just in case code is copied/pasted around :)

apoudret-microej: that's a good reason.
fixed (amend)

gjadi-mej (approved):

PR #1065, opened 2024-10-07
Clarify usage of LLMJVM_on_CheckIntegrity_error

jerome-leroux:


VEEPortingGuide/coreEngine.rst
```diff
@@ -584,20 +584,30 @@ The following internal structures may be modified without affecting the checksum
...
-.. code-block:: java
+.. code-block:: c
 
-   void Java_com_mycompany_MyClass_myNativeFunction(void) {
-   		int32_t crcBefore = LLMJVM_checkIntegrity();
-   		myNativeFunctionDo();
+    #include <stdio.h>
+    #include "LLMJVM.h"
+    
+    void LLMJVM_on_CheckIntegrity_error(uint32_t errorCode, void* errorAddress);
```
frivieremicroej, line 595: why declaring the prototype  ? (it is not called by the user code). declaring the hook body below is enough.

jerome-leroux: This may be required with some compiler options.

jerome-leroux: Actually, the function is not referenced by the user code, so we don't need to declare it before its definition.

frivieremicroej (approved):

PR #1070, opened 2024-10-11
document how to create custom ExecTool tasks

denisbedin:
https://docs.microej.com/en/feature-sdk6_create_custom_exectool_tasks/SDK6UserGuide/howtos/createCustomExecToolTask.html

### DIFF
--- a/SDK6UserGuide/howtos/createCustomExecToolTask.rst
+++ b/SDK6UserGuide/howtos/createCustomExecToolTask.rst
@@ -1,0 +1,107 @@
+.. _sdk_6_create_cutom_exectool_task:
+
+How To Create Custom ExecTool Task
+==================================
+
+MicroEJ provides the Gradle ``execTool`` task that allows to execute all :ref:`Command Line Tools <sdk6_microejtools>`. 
+However, it is possible to create a custom task of type ``ExecToolTask`` specific to a tool. This task can then be used to execute the tool before or after another task when 
+building your project for example.
+
+You can create a new ``ExecToolTask`` in the build file of your project as follows:
+
+- Import the ``ExecToolTask`` and ``LoadVeeTask`` classes::
+
+    import com.microej.gradle.tasks.ExecToolTask
+    import com.microej.gradle.tasks.LoadVeeTask
+
+- Create a new task of type ``ExecToolTask``::
+
+    val loadVee = tasks.withType(LoadVeeTask::class).named("loadVee")
+    val mainSourceSet = project.extensions.getByType(SourceSetContainer::class).getByName(SourceSet.MAIN_SOURCE_SET_NAME)
+
+    tasks.register<ExecToolTask>("myCustomTask") {
+        veeDir.set(loadVee.flatMap { it.loadedVeeDir })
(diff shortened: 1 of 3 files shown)


SDK6UserGuide/howtos/createCustomExecToolTask.rst
```diff
@@ -0,0 +1,102 @@
+.. _sdk_6_create_cutom_exectool_task:
+
+How To Create Custom ExecTool Task
+==================================
+
+MicroEJ provides the Gradle ``execTool`` task that allows to execute :ref:`Command Line Tools <sdk6_microejtools>`. This task is generic and works with all MicroEJ tools,
+but it is possible to create a custom Gradle task for a specific tool as follows:
```
tdelhomenie-microej, line 7: We should add an use case example, why would I need to create a task for a specific tool ? The use case we met is for automatically executing the tool after or before another task in the build.

SDK6UserGuide/howtos/createCustomExecToolTask.rst
```diff
@@ -0,0 +1,102 @@
...
+    These inputs are required for all tasks of type ``ExecToolTask``.
+
+- Set the name of the Tool and the required properties::
+
+    toolName = "microej_tool_name"
+    toolProperties.putAll(mapOf(
+        "property1" to "value1",
+        "property2" to "value2",
+        ...
+    ))
```
tdelhomenie-microej, line 37: Where should I put these lines? If they must be put in the task definition, it should be moved directly in the previous code snippet.

SDK6UserGuide/howtos/createCustomExecToolTask.rst
```diff
@@ -0,0 +1,102 @@
...
+
+- The custom task can then be executed::
+
+    $ ./gradlew myCustomTask
+
+As an example, here is the code of a custom Gradle task specific to the :ref:`sdk6_localDeployTool`::
```
tdelhomenie-microej, line 43: We should describe what the example does exactly, something like "Here is how create a task which calls the local deploy tool and ...".
And we could add the code to execute it after the buildFeature task for example, that would be a complete example. WDYT?

tdelhomenie-microej (changes requested):

SDK6UserGuide/howtos/createCustomExecToolTask.rst
denisbedin: Done.

SDK6UserGuide/howtos/createCustomExecToolTask.rst
denisbedin: I moved them in the other code snippet. Actually, I first wanted to separate the inputs common to all tasks of type "ExecToolTask" and the inputs specific to the tool, but it was not very clear so I fixed it.

SDK6UserGuide/howtos/createCustomExecToolTask.rst
denisbedin: I fixed the sentence, I didn't added the example with "buildFeature" since, as discussed, the .fo will be built two times.

tdelhomenie-microej (approved):

PR #1104, opened 2024-11-08
Clarify usage and impacts of microejVee configuration in dependencies

tdelhomenie-microej:


rlouedec-microej:
I think what was confusing for the customer is not that the VEE Port dependency could add libraries transitively.
I think it was that a VEE Port dependency configured on a library would be transitively used by applications depending on this library.

rlouedec-microej (changes requested):
See https://github.com/MicroEJ/docs/pull/1104#issuecomment-2464797245

SDK6UserGuide/testProject.rst
```diff
@@ -57,6 +57,10 @@ Executing tests on the Simulator allows to check the behavior of the code in an
 but without requiring the board.
 This solution is therefore less constraining and more portable than testing on the board.
 
+The target VEE Port must be declared in the projects dependencies, as explained in :ref:`sdk_6_select_veeport`.
+Declaring a VEE Port in project dependencies does not fetch any transitive libraries,
+it is only meant to use the declared VEE Port for all the operations requiring a VEE Port, 
+including testing a project on Simulator or on Device.
```
rlouedec-microej, line 63: Maybe something like:
```suggestion
The target VEE Port must be declared in the projects dependencies, as explained in :ref:`sdk_6_select_veeport`.
Declaring a VEE Port in project dependencies only applies to the current project. This configuration is not fetched transitively by dependent projects.
Especially when configuring the VEE Port to test a library projects, application projects depending on this library will not "see" this test VEE Port, must configure a VEE Port on their own and are free to use a different one.
```

tdelhomenie-microej: Ok, indeed I understood the main concern. I applied your suggestion with a small change ("dependent" -> "consumer"), I let you check.

SDK6UserGuide/selectVeePort.rst
```diff
@@ -7,6 +7,14 @@ Building or running an Application or a Test Suite with the SDK requires a VEE P
 
 Use one of the following available options to provide it to your project. 
 
+.. note::
+
+   Declaring a VEE Port in project dependencies only applies to the current project. 
+   This configuration is not fetched transitively by consumer projects.
+   Especially when configuring the VEE Port to test a library projects,
```
rlouedec-microej, line 14: That's my bad but:
```suggestion
   Especially when configuring the VEE Port to test a library project, 
```

SDK6UserGuide/testProject.rst
```diff
@@ -57,6 +57,13 @@ Executing tests on the Simulator allows to check the behavior of the code in an
 but without requiring the board.
 This solution is therefore less constraining and more portable than testing on the board.
 
+The target VEE Port must be declared in the projects dependencies, as explained in :ref:`sdk_6_select_veeport`.
+
+Declaring a VEE Port in project dependencies only applies to the current project. 
+This configuration is not fetched transitively by consumer projects.
+Especially when configuring the VEE Port to test a library projects,
```
rlouedec-microej, line 64: That's my bad but:
```suggestion
Especially when configuring the VEE Port to test a library project, 
```

rlouedec-microej (changes requested):
LGTM, just typo

SDK6UserGuide/testProject.rst
rlouedec-microej: OK 👍

SDK6UserGuide/testProject.rst
tdelhomenie-microej: Done.

SDK6UserGuide/selectVeePort.rst
tdelhomenie-microej: Done.

PR #1109, opened 2024-11-13
Retell that createVM and destroyVM should be called only once

gjadi-mej:
Even though it's explained above, when I looked at how to restart the core engine, I only looked at the `restart core engine` section, not the whole page.


VEEPortingGuide/coreEngine.rst
```diff
@@ -432,7 +432,12 @@ To restart the application, call again the ``SNI_startVM`` function (see the fol
 
    Please note that while the Core Engine supports restart, :ref:`MicroUI <section_microui>` does not. 
    Attempting to restart the Application on a VEE Port with UI support may result in undefined behavior.
-     
+
+
+.. note::
+
+   Please note that `SNI_createVM`` and ``SNI_destroyVM`` should only be called once.
```
jerome-leroux, line 439:
```suggestion
   Please note that ``SNI_createVM`` and ``SNI_destroyVM`` should only be called once. 
```

jerome-leroux (approved):

PR #1115, opened 2024-11-19
Comment option for trace redirection by default

tdelhomenie-microej:


SDK6UserGuide/testProject.rst
```diff
@@ -317,11 +317,12 @@ The configuration is similar to the one used to execute a testsuite on the Simul
                               // Enable the build of the Executable
                               "microej.testsuite.properties.deploy.bsp.microejscript" to "true",
                               "microej.testsuite.properties.microejtool.deploy.name" to "deployToolBSPRun",
-                              // Tell the testsuite engine that the VEE Port Run script redirects execution traces
-                              "microej.testsuite.properties.launch.test.trace.file" to "true",
                               // Configure the TCP/IP address and port if the VEE Port Run script does not redirect execution traces
                               "microej.testsuite.properties.testsuite.trace.ip" to "localhost",
-                              "microej.testsuite.properties.testsuite.trace.port" to "5555"
+                              "microej.testsuite.properties.testsuite.trace.port" to "5555",
```
bguedas, line 322: Is the comma useful? By default, the other lines are commented. The other option is to uncoment the last line but comment the line with the comma...

SDK6UserGuide/testProject.rst
```diff
@@ -317,11 +317,12 @@ The configuration is similar to the one used to execute a testsuite on the Simul
                               // Enable the build of the Executable
                               "microej.testsuite.properties.deploy.bsp.microejscript" to "true",
                               "microej.testsuite.properties.microejtool.deploy.name" to "deployToolBSPRun",
-                              // Tell the testsuite engine that the VEE Port Run script redirects execution traces
```
bguedas, line 320: Can you add newlines to separate blocks (comment + option)?

bguedas (changes requested):

SDK6UserGuide/testProject.rst
tdelhomenie-microej: Having a trailing comma is supported, and allow the configuration to still be valid if the developer uncomment the last line but forget to comment the port and ip properties lines, but I would keep it like that.

SDK6UserGuide/testProject.rst
tdelhomenie-microej: I added a blank line before the "Configure the TCP/IP address ..." comment, but kept the other options together since they are very related.

bguedas (approved):

PR #1125, opened 2024-11-26
Update SNI specification to allow non-immortal arrays usage.

jerome-leroux:


ApplicationDeveloperGuide/runtime.rst
```diff
@@ -79,7 +79,7 @@ SNI provides a simple mechanism for implementing native Java methods in the C la
 SNI allows you to:
 
 -  Call a C function from a Java method.
--  Access an Immortal array in a C function (see the :ref:`runtime_bon` to learn about immortal objects).
+-  Share memory between C and Java using Immortal arrays (see the :ref:`runtime_bon` to learn about immortal objects).
```
frivieremicroej, line 82: So we should relax immortal arrays here ?

frivieremicroej (changes requested):

ApplicationDeveloperGuide/runtime.rst
jerome-leroux: The real shared memory mechanism  (between java, c tasks, interrupt, etc.) requires immortal arrays

frivieremicroej: OK

frivieremicroej (approved):

PR #1152, opened 2024-12-18
Explicit Sphinx conf as required by RTD

gharelMicroEJ:


.readthedocs.yml
```diff
@@ -12,6 +12,7 @@ formats:
 
 sphinx:
```
frivieremicroej, line 13: Shall we declare the version as mentionned in RTD email ?

```suggestion
version: 2
sphinx:
```

gharelMicroEJ: The version is already at the top of the file.

frivieremicroej (approved):

PR #1170, opened 2025-01-10
document how to call Java from ManagedC by referencing the java class…

agazardMicroeJ:
… by its module name


frivieremicroej:
Overall Approved by #1184

PR #1174, opened 2025-01-14
Fix getting started sdk 6

frivieremicroej:


rbabaousmail:
I think all is good, just the copyright year was not updated

PR #1196, opened 2025-01-29
Update the number of thread limit in multi-sandbox mode (increased to 255)

agazardMicroeJ:


jerome-leroux (approved):

jerome-leroux (commented):
Should we specify from which Architecture version the change has been done?

jerome-leroux (approved):

PR #1219, opened 2025-02-18
Clarify External Resources Folder usage

acolleux-microej:


VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
 .. _external_resources_folder:
 
-External Resources Folder
-=========================
+Pre-Processed External Resources
+================================
+
+When working with :ref:`chapter.microej.applicationResources` declared as external resources, the Application build process will
+output those external resources in a dedicated output folder named ``externalResources/``.
+This folder gathers all the resources that should be programmed on the device.
```
rlouedec-microej, line 86:
```suggestion
This folder gathers all the resources that should be deployed on the device.
```

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
 
 .. _external_resources_folder:
 
-External Resources Folder
-=========================
+Pre-Processed External Resources
+================================
+
+When working with :ref:`chapter.microej.applicationResources` declared as external resources, the Application build process will
+output those external resources in a dedicated output folder named ``externalResources/``.
```
rlouedec-microej, line 85: I would think the idea of this section is to discuss how to get the external resource files that are pre-processed from the SDK in order to deploy them on the device.

Here you referring to all external resources in general. I would suggest:
```suggestion
Some :ref:`chapter.microej.applicationResources` are intended to be pre-processed by the MICROEJ SDK (typically formatted to optimize footprint and/or parsing/processing).
That is for example the case of MicroUI :ref:`section.ui.Images` & :ref:`section.ui.Fonts` as well as :ref:`chapter.nls`.
When resources are marked as external, they are pre-processed during the application build and available in the build output folder:
```

And I would have the "tabs" directly after this. WDYT?

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
+The location of the ``externalResources/`` folder is different between SDK 5 and SDK 6:
+
+.. tabs::
+
+   .. tab:: SDK 6
+
+      The ``externalResources/`` folder is located in the ``build/application/object`` folder of the application project.
+
+      It is generated when building the application for the device (:ref:`sdk6_module_natures.tasks.buildExecutable`, :ref:`sdk6_module_natures.tasks.runOnDevice`
+      or :ref:`sdk6_module_natures.tasks.buildApplicationObjectFile`).
```
rlouedec-microej, line 100: That's too much details IMO, we could add the standard Gradle build task (or even publish, why not?).
If we want to be precise I would only mention the task that actually produces this output folder (which I assume is `buildApplicationObjectFile`).
I suggest to be precise here and assume we have educated our users to always use the standard tasks (build / test / publish):
```suggestion
      It is generated when building the application for the device (:ref:`sdk6_module_natures.tasks.buildApplicationObjectFile`).
```

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
+      (e.g. ``Example-ExternalResourceLoader/com.microej.externalresourceloader.ExternalImages/externalResources``).
+
+Simulation
+==========
 
-The External Resource Loader module provides an option (MicroEJ launcher
-option) to specify a folder for the external resources. This folder has
-two roles:
+The :ref:`chapter.microej.applicationResources` declared as external resources are automatically taken into account
+by the Simulator, thanks to their declarations in ``*ext.list`` files.
```
rlouedec-microej, line 112: I suggest to reword:
```suggestion
The :ref:`chapter.microej.applicationResources` provided by the application project, including those marked as external resources, and those that need pre-processing, are made available to the application during the simulation.
```
If we want to further clarify, we could add:
```
There is no need to define an external resource loader for the simulator.
```

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
 
--  It is the output folder used by some extra generators during the
-   MicroEJ Application build. All output files generated by these tools
-   will be copied into this folder. This makes it easier to retrieve the
-   exhaustive list of resources to program on the board.
+However, it is also possible to provide an input folder containing all the External Resources of the Application.
```
rlouedec-microej, line 114: Here it is unclear, does it override the existing mechanism (external resources provided by the application project, & dependencies)?

Or does it add to it?

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
 
--  This folder is taken into consideration by the Simulator in order to
-   simulate the availability of these resources. When the resources are
-   located in another computer folder, the Simulator is not able to load
-   them.
+That way, the Simulator will simulate the availability of these resources.
+If they are not made available in this folder, the Simulator will not able to load them.
+This can be useful to mimic the resources made available on the device (e.g. share an SD Card between the Simulator and the device containing the same External Resources)
```
rlouedec-microej, line 118: Oh nice use case!

"That way, the Simulator will simulate the availability of these resources." => it is not really "simulation" here, the resources are really available. (what we are simulating is the external resource loader mechanism)

"If they are not made available in this folder, the Simulator will not **be** able to load them." => does it mean the configurable overrides the default mechanism like I discussed in my previous comment?

I would clarify previous sentence and keep only this here:
```suggestion
This can be useful to mimic the external resources storage on the device (e.g. share an SD Card between the Simulator and the device).
```

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -78,27 +78,49 @@ APIs such as ``read``, ``mark`` etc.
...
 
-If not specified, this folder is created (if it does not already exist)
-in the MicroEJ project specified in the MicroEJ launcher. Its name is
-``externalResources``.
+Follow the steps below to provide an input folder containing External Resources:
 
+- In the Application project, rename all the ``*ext.list`` files to ``*ext.list.tmp``.
```
rlouedec-microej, line 122: Why?

Also, is this really working? I would assume not for images/fonts given the MicroUI implementation relies on having a list of them AFAIK.

acolleux-microej: Ok, actually I think it's a bug, I think we shouldn't document this feature while it's not fixed.

It looks like the simulator is looking into its classpath and first uses the images from the resources folder before the one declared as external resources folder:

image

image

VEEPortingGuide/externalResourceLoader.rst
acolleux-microej: Cf comment below

VEEPortingGuide/externalResourceLoader.rst
acolleux-microej: The External Resources Folder stores processed resources and also unprocessed resources. Is there a point of distinguishing the 2 ?

image

rlouedec-microej: I get your point.
My point is why telling the user how to get access to resources he already has?
And I think it is confusing to put this in the "preprocessed" section.

But then, if we apply my suggestions, maybe we can add the following at the end:
```
The typical development flow is:

1. Build the application & program your device.
2. Test your application with the external resources (it is possible to change them after the executable build).
3. When good, re-build the application, and collect the external resources in the output folder

   .. note::

      The output folder not contains the pre-processed resources but also all the other external resources from the project that are not pre-processed.

4. Deploy the external resources on your device.

File: VEEPortingGuide/externalResourceLoader.rst (hunk 7 above)
Comment (rlouedec-microej) [reply VEEPortingGuide/externalResourceLoader.rst:118]: Well, first, it seems you are integrating `robot.png` as a MicroUI image.
Therefore it should be preprocessed (the actual external resource name should be `robot.png_raw`).
I would assume this preprocessed image should be the one to put in `ExternalImages/externalResources/images` (and not robot.png).
My expectation is that doing that instead in your first step would fix your issue.

Also, my expectation would be that if you remove the image from the list, then the image should not be available to the application and you should have an error (and I quite don't get how you can possibly have this output in the 2nd scenario => let's investigate).

Comment (rlouedec-microej) [reply VEEPortingGuide/externalResourceLoader.rst:118]: Actually, I think I got it wrong: you are using the built-in runtime PNG decoder here..
In that case, I agree with you on the expectations for scenario 1.
For scenario 2, I am not sure, it may depend on the version of the UI pack, I know this was not supported at least for fonts some time ago but this may have been improved recently.

Comment (acolleux-microej) [reply VEEPortingGuide/externalResourceLoader.rst:118]: Yes, I'm using the PNG decoder. I'm using the latest 14.3 pack

File: VEEPortingGuide/externalResourceLoader.rst (hunk 2 above)
Comment (acolleux-microej) [reply VEEPortingGuide/externalResourceLoader.rst:85]: I think it's convenient for the user to have a single output folder to deploy all of its resources at once.
The user shouldn't have to sort its resources and know which are the ones that should be taken in the /resources/ folder (e.g. PNG) and which should be taken in the externalResources/ folder.

Comment (acolleux-microej) [reply VEEPortingGuide/externalResourceLoader.rst:85]: On second thoughts, we should maybe keep the title as before (External Resources Folder). Here is what I suggest:

```
External Resources Folder
=========================

When working with :ref:`chapter.microej.applicationResources` declared as external resources, the Application build process will
output those external resources in a dedicated output folder named ``externalResources/``.

This folder gathers all the :ref:`chapter.microej.applicationResources` that should be deployed on the device.
This folder not only contains the pre-processed resources but also all the other external resources from the project that are not pre-processed:

- Pre-processed resources: resources formatted by MICROEJ SDK to optimize footprint and/or parsing/processing. 
  That is for example the case of :ref:`MicroUI Fonts <section.ui.Fonts>`, :ref:`NLS Resources <chapter.nls>` and :ref:`MicroUI Images embedded in RAW format <section.ui.Images>`.
- Unprocessed resources: resources embedded as-is, like :ref:`Raw Resources <section.classpath.elements.raw_resources>` or 
  :ref:`section.ui.Images` with no output format defined (e.g. image embedded as a ``.png``).

The location of the ``externalResources/`` folder is different between SDK 5 and SDK 6:

.. tabs::

   .. tab:: SDK 6

      The ``externalResources/`` folder is located in the ``build/application/object`` folder of the application project.

      It is generated when building the application for the device (:ref:`sdk6_module_natures.tasks.buildApplicationObjectFile`).

   .. tab:: SDK 5

      The ``externalResources/`` folder is located in the output folder of the application project.
      This folder is defined in the ``Execution`` tab of the :ref:`concepts-microejlaunches` configuration
      (e.g. ``Example-ExternalResourceLoader/com.microej.externalresourceloader.ExternalImages/externalResources``).


Development Flow
================

The typical development flow when working with External Resources on the device is:

1. Build the application & program your device,
2. Test your application with the external resources (it is possible to change them after the executable build),
3. When good, re-build the application, and collect the external resources in the output folder,
4. Deploy the external resources on your device.

```

File: VEEPortingGuide/externalResourceLoader.rst [hunk 8]
```diff
@@ -132,6 +140,28 @@ Use
...
 Application tries to open an external resource.
 
+A simple implementation of the External Resources Loader is available on GitHub:
+`Example-ExternalResourceLoader <https://github.com/MicroEJ/Example-ExternalResourceLoader>`_.
+
+Simulation
+----------
+
+The :ref:`chapter.microej.applicationResources` provided by the application project, 
+including those marked as external resources, are automatically
```
rlouedec-microej, line 150: Update according to changes in other sections:
```suggestion
The :ref:`chapter.microej.applicationResources` provided by the application project, 
pre-processed and/or unprocessed, are automatically 
```

VEEPortingGuide/externalResourceLoader.rst
rlouedec-microej: Great!

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -132,6 +140,28 @@ Use
...
+
+The :ref:`chapter.microej.applicationResources` provided by the application project, 
+including those marked as external resources, are automatically 
+made available to the application during the simulation.
+
+On Device 
+---------
+
+The :ref:`chapter.microej.applicationResources` marked as external resources must be deployed
+on the device before running the application.
```
rlouedec-microej, line 157: Technically, they could be deployed after the application is started:
```suggestion
The external resources must be deployed on the device before they can be consumed by the application.
```

rlouedec-microej (approved):

VEEPortingGuide/externalResourceLoader.rst
```diff
@@ -132,6 +140,28 @@ Use
...
+
+The typical development flow when working with External Resources on the device is:
+
+1. Build the application & program the device,
+2. Test the application with the external resources (it is possible to change them after the executable build),
+3. When good, re-build the application, and collect the external resources in the output folder,
+4. Deploy the external resources on your device (e.g. copy them to the device FileSystem).
```
rlouedec-microej, line 164:
```suggestion
After that, the external resources can be updated (without re-building the application) with the following development flow:

1. Update the external resources in the application project.
2. Test the application with the updated resources on the simulator.
3. When good, re-build the application, and collect the external resources in the output folder (the build is required to trigger the processing of the resources).
4. Deploy the external resources on the device (e.g. copy them to the file system of the device: removable SD card, USB mass storage, ...).
```

PR #1225, opened 2025-02-25
Add Note to explain how to increase HIL maximum frame size

cblanchard-microej:
https://docs.microej.com/en/feature-hil_frame_size/VEEPortingGuide/mock.html#mocks-design-support

### DIFF
--- a/VEEPortingGuide/mock.rst
+++ b/VEEPortingGuide/mock.rst
@@ -145,12 +145,6 @@ Interface
 The MicroEJ Simulator interface is defined by static methods on the Java
 class ``com.is2t.hil.NativeInterface``.
 
-.. note::
-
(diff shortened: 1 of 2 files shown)


VEEPortingGuide/mock.rst
```diff
@@ -145,6 +145,12 @@ Interface
 The MicroEJ Simulator interface is defined by static methods on the Java
 class ``com.is2t.hil.NativeInterface``.
 
+.. note::
```
frivieremicroej, line 148: Move the note to `Resource Management` section

frivieremicroej (changes requested):

VEEPortingGuide/mock.rst
cblanchard-microej: Are you sure is is specific to resource Management? What if we want to pass a big array?

frivieremicroej (approved):

PR #1226, opened 2025-02-25
Update documentation about hashcode.

seonmej:


Trainings/tutorialImproveCodeQuality.rst
```diff
@@ -338,17 +338,23 @@ The ``equals(Object)`` method is written that way:
...
 
 - Choose a prime number.
-- Create a result local, whatever the value (usually the prime number).
-- For each field, multiply the previous result with the prime
-  plus the hash code of the field and store it as the result.
+- Create a result local intialized with the hashcode of the most significant field.
+- For each remaining field, multiply the previous result with the prime plus the hash code of the field and store it as the result.
 - Return the result.
+- Only the fields used in ``equals()`` must be used.
+- The fields computed from other fields can be ignored.
```
apoudret-microej, line 351: `The fields computed from other fields`? you meant the value of the other fields?

Trainings/tutorialImproveCodeQuality.rst
```diff
@@ -338,17 +338,23 @@ The ``equals(Object)`` method is written that way:
       (field2 == null ? other.field2 == null : field2.equals(other.field2));
   }
 
-The `Object.hashCode()`_ method is written that way:
+The goal of the `Object.hashCode()`_ is to produce different values for unequals objects.
```
apoudret-microej, line 341:
```suggestion
The goal of the `Object.hashCode()`_ is to produce different values for unequal objects.
```

Trainings/tutorialImproveCodeQuality.rst
```diff
@@ -338,17 +338,23 @@ The ``equals(Object)`` method is written that way:
...
+- Only the fields used in ``equals()`` must be used.
+- The fields computed from other fields can be ignored.
+- Precomputing the hashcode may be convenient for performance purpose (especially when fields are final).
+- The hashcode can also be lazy initialized the first time it is requested.
 
 Depending on its type, the hash code of a field is:
 
-- Boolean: ``(f ? 0 : 1)``.
+- Boolean: ``(f ? -1 : 1)``. Or choose two different prime numbers instead of ``-1`` and ``1``.
 - Byte, char, short, int: ``(int) f)``.
```
apoudret-microej, line 358:
```suggestion
- Byte, char, short, int: ``(int) f``.
```

Trainings/tutorialImproveCodeQuality.rst
```diff
@@ -338,17 +338,23 @@ The ``equals(Object)`` method is written that way:
       (field2 == null ? other.field2 == null : field2.equals(other.field2));
   }
 
-The `Object.hashCode()`_ method is written that way:
+The goal of the `Object.hashCode()`_ is to produce different values for unequals objects.
+A good hashcode is uniformly distributed among hash buckets (for instance in HashMap, HashSet, etc.)
+
+The ``hashCode()`` method is written that way:
 
 - Choose a prime number.
```
apoudret-microej, line 346: could you provide recommendations on which prime number to choose? does it depend on the value range of fields, or from the number of fields?

Trainings/tutorialImproveCodeQuality.rst
```diff
@@ -338,17 +338,23 @@ The ``equals(Object)`` method is written that way:
...
 - Return the result.
+- Only the fields used in ``equals()`` must be used.
+- The fields computed from other fields can be ignored.
+- Precomputing the hashcode may be convenient for performance purpose (especially when fields are final).
+- The hashcode can also be lazy initialized the first time it is requested.
 
 Depending on its type, the hash code of a field is:
 
-- Boolean: ``(f ? 0 : 1)``.
+- Boolean: ``(f ? -1 : 1)``. Or choose two different prime numbers instead of ``-1`` and ``1``.
```
apoudret-microej, line 357: can you explain this change? effective java (https://thefinestartist.com/effective-java/09) recommends 1 and 0

seonmej: You mention the older version of Effective Java. The [new one](https://ahdak.github.io/blog/effective-java-part-2/) recommends to use the static `hashCode()` methods from the Boxed type. For instance `Boolean.hashCode(boolean)` is not available in EDC but is implemented `f ? 1231 : 1237;` in open JDK. We can also propose that? Otherwise `-1` seems a better multiplier than `0` to distribute the result value.

Trainings/tutorialImproveCodeQuality.rst
seonmej: It depends on the number of hash buckets. 31 seems (from what I have read) a large enough prime that the number of buckets is unlikely to be divisible by it.

Trainings/tutorialImproveCodeQuality.rst
apoudret-microej: OK, Oracle JDK also uses 1231 and 1237 so I think we can suggest these numbers.
=> Easy to understand/maintain solution: 1 and 0. Optimal solution to minimize collisions: 1231 and 1237.

Trainings/tutorialImproveCodeQuality.rst
apoudret-microej: can we suggest 31 in the doc?

seonmej: it's already done ;-)

Trainings/tutorialImproveCodeQuality.rst
seonmej: I just put the solution done in the JDKs.

PR #1237, opened 2025-03-05
Fix code sample typo

stinlot-microej:


stinlot-microej:
Also im not sure but at line: https://github.com/MicroEJ/docs/blob/066a20d5c088102e17ba08f99ad3cf5419fd5bda/ApplicationDeveloperGuide/managedc/memory.rst?plain=1#L97

Isn't it supposed to be 512 instead of 612 ?

frivieremicroej (approved):

PR #1243, opened 2025-03-07
[WIP] Reword & clarify transmit & swap buffer policy / predraw BTS

rlouedec-microej:


VEEPortingGuide/uiDisplay.rst
```diff
@@ -365,6 +368,18 @@ The *flush* step consists in copying the back buffer data to the front buffer **
 
    Transmit and Swap (parallel)
 
+.. note::
+
+   To render the next frame on a different buffer, it may be needed to first restore the content of the new active back buffer.
+   This is described in :ref:`section_brs`. This "Transmit & Swap (parallel)" policy is best combined with the :ref:`section_brs_predraw`.
+   See details on the optimizations in :ref:`Predraw applied to Transmit & Swap <section_brs_predraw_transmitswap>`.`
+
+.. I am not sure to understand here: to restore, we would need to perform a copy between the two back buffers. This can only be
```
gbalan-microej, line 377: You're right but this chapter does not mention the "restore": the idea is to describe the different techs of buffering. Don't care about restoration. The choose between "serial" and "parallel" mode is an hardware restriction. Then, according to the available memory, you can use 1 or 2 buffers in RAM in addition with the "serial" buffer (on LCD side). In most of cases, the "serial" mode is slow: the transmit is slower than a copy RAM to RAM.

VEEPortingGuide/uiDisplay.rst
```diff
@@ -344,19 +342,24 @@ The *flush* step consists in transmitting the back buffer data to the display mo
 
    Transmit and Swap (serial)
 
+.. note::
+
+   To render the next frame on a different buffer, it may be needed to first restore the content of the new active back buffer.
+   This is described in :ref:`section_brs`. This "Transmit & Swap (serial)" policy is best combined with the :ref:`section_brs_predraw`.
+   See details on the optimizations in :ref:`Predraw applied to Transmit & Swap <section_brs_predraw_transmitswap>`.`
```
gbalan-microej, line 349: As I said in another comment, this chapter does not mention the problem of restore (it is not the aim of this chapter). But I understand why you mention it here. I agree to keep your note but this note should be copied here: Swap Double Buffer (parallel) and Swap Triple Buffer (parallel),

VEEPortingGuide/uiBRS.rst
```diff
@@ -363,23 +363,29 @@ Strategy: Predraw
 Principle
 ---------
 
-This strategy considers that the drawings are always performed in a buffer, and a swap with another buffer is made by the implementation of ``LLUI_DISPLAY_IMPL_flush()``.
-In this case, the restoration is mandatory because the new back buffer must contain the past before the buffer swapping.
+This strategy can be used to refresh an outdated buffer to the active back buffer before they are swapped.
```
gbalan-microej, line 366: The chapter pattern is the same for all BRS chapter. If you change one, change all

gbalan-microej: Your sentence is wrong or I don't understand what you want to say. The BB is first restored before the first drawing. It is not restored before the swap

VEEPortingGuide/uiBRS.rst
```diff
@@ -363,23 +363,29 @@ Strategy: Predraw
...
 
 Behavior
 --------
 
 The following table illustrates how the strategy works:
 
+.. I wonder if we could improve this by adding the actual API calls `Display.flush()`, `MyPainter.drawA()`,
+   `MyPainter.drawB()`, `MyPainter.drawC()`, ..
+   But I am not sure how to do that with a simple table (one call corresponds to multiple actions)
+
```
gbalan-microej, line 388: We don't care about the java code: here the LLAPI code is enough according to me (vee port guy). Maybe we can add another chapter that describes the calls java -> llapi -> driver to not overload this table (which is enough complex to understand)

VEEPortingGuide/uiBRS.rst
```diff
@@ -388,7 +394,7 @@ The following table illustrates how the strategy works:
      - Strategy Work
      - Back Buffer  
      - Front Buffer 
-   * - Startup
+   * - Initial State
```
gbalan-microej, line 397: Ok for this change but must be copied everywhere

VEEPortingGuide/uiBRS.rst
```diff
@@ -540,29 +548,23 @@ Here are the steps around the strategy describing how to use it in :ref:`transmi
 
         Transmit and Swap (parallel)
 
-The two buffers have the same role alternatively: back buffer and transmission buffer.
-On startup, the transmission buffer has yet to be used.
+In this policy, the newly active back buffer is refreshed after a swap.
```
gbalan-microej, line 551: cf above: I think you have not understand when the BB is updated... maybe we can discuss together because your approach seems better.

VEEPortingGuide/uiBRS.rst
```diff
@@ -540,29 +548,23 @@ Here are the steps around the strategy describing how to use it in :ref:`transmi
...
+   a. If the Application requests a new drawing before the frame transfer is completed, the "pre-draw" refresh is performed.
+   b. If the frame transfer is completed before the Application requests a new drawing, the buffers are "swapped" back (the pointer to "current buffer for drawing" is reverted to buffer A).
+      In that case, no refresh is needed, and the Application can continue drawing on the same back buffer. In this case, the regions to be refreshed are cumulates the drawing of multiple frames.
 
-14.  The BSP should notify the Graphics Engine again by calling ``LLUI_DISPLAY_setBackBuffer()``, giving the transmission buffer address: the Graphics Engine will reuse this buffer for future drawings, and the strategy will not need to restore anything.
+.. I understand that I am removing some (maybe useful) documentation on the details of the sequence of native API calls.
+   I think this helps to clarify the concept. We may add this back in the typical implementations section?
```
gbalan-microej, line 567: not agreed to remove all steps. These are the steps that occur between the driver, the interrupt, the graphics engine, etc.

VEEPortingGuide/uiDisplay.rst
rlouedec-microej: Sorry, I think I meant "refresh", not "restore".

> In most of cases, the "serial" mode is slow: the transmit is slower than a copy RAM to RAM.

That's kind of my point: in "parallel" mode, why would you use this policy? It seems inefficient: uses more memory than the single buffer policy (3 instead of 2) and has no benefit in terms of performance (it can even lower the performances with additional useless refresh/copies).

VEEPortingGuide/uiDisplay.rst
rlouedec-microej: I agree this should be applied to other policies 👍 
Also, reword "restore" to "refresh".

VEEPortingGuide/uiDisplay.rst
gbalan-microej: parallel 3: useful for the RT1170! :)

VEEPortingGuide/uiBRS.rst
rlouedec-microej: Well obviously we have different perspectives here. As a Java developer, I would be interested with having this extra info.
Main reason I think was that, with just the current table, it was not clear what triggered the implicit region calls. (after looking more into it, I think I got it but it's not obvious.)
Anyway, I am not sure if I would have a better way to present this including the high level API calls.

VEEPortingGuide/uiBRS.rst
rlouedec-microej: From my understanding, it is refreshed after a swap, just before the first drawing.
Sure, we can discuss that.

VEEPortingGuide/uiDisplay.rst
rlouedec-microej: Using "Transmit & Swap (parallel)" or "Swap double (parallel)"?

PR #1266, opened 2025-03-31
Add a Mock from an Application

tdelhomenie-microej:


VEEPortingGuide/mock.rst
```diff
@@ -365,6 +368,38 @@ Installation
...
+You can also install a Mock from an Application project, for example when a native function is added directly in the Application project.
+When a Mock is declared in an Application, it is automatically added in the VEE Port used to build or run the Application.
+
+.. tabs::
+
+   .. tab:: SDK 6
+
+      .. warning::
+
+         A Mock can be installed from an SDK 6 Application project only if the VEE Port used in the project is an SDK 6 VEE Port.
```
bguedas, line 383: How do we now that the VEE Port is SDK 6 or SDK 5? Shouldn't we give a way to check it?

bguedas (approved):

VEEPortingGuide/mock.rst
bguedas: Let's see if someone gets this case.

PR #1267, opened 2025-04-02
Document how to provide a VEE dedicated to a testsuite

denisbedin:


SDK6UserGuide/testProject.rst
```diff
@@ -629,6 +629,26 @@ since it avoids adding confguration to change the tests sources folder.
 With this configuration, tests on the Simulator are located in the ``src/test/java`` folder, 
 and tests on a Java SE VM are located in the ``src/testOnJavaSe/java`` folder.
 
+.. _sdk_6_vee_configuration_by_testsuite:
+
+Configure a VEE by Testsuite
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
tdelhomenie-microej, line 635: We already have a section for the VEE configuration (https://docs.microej.com/en/latest/SDK6UserGuide/testProject.html#configure-the-vee). I agree it should not be in the same section, since it is an advanced use case, but it is also weird to have it in another section at the same level.
Shoudn't we have a "Advanced" section, or something like that, to gather these use cases?

SDK6UserGuide/testProject.rst
```diff
@@ -629,6 +629,26 @@ since it avoids adding confguration to change the tests sources folder.
...
+
+The VEE declared in the project dependencies with the ``microejVee`` or the ``testMicroejVee`` configuration 
+(refer to :ref:`sdk_6_testsuite_vee_configuration` for more details) is used to execute all testsuites, 
+but it is also possible to run a testsuite on a dedicated VEE. To define a VEE for your testsuite, you must:
```
tdelhomenie-microej, line 639: It should be clear that it is for testsuites other than the default `test` testsuite.

SDK6UserGuide/testProject.rst
denisbedin: I moved it to a new "Test Suite Advanced Configuration" section.

SDK6UserGuide/testProject.rst
denisbedin: Done.

SDK6UserGuide/testProject.rst
```diff
@@ -629,6 +629,26 @@ since it avoids adding confguration to change the tests sources folder.
...
+- Create a new ``<testsuite_name>MicroejVee`` configuration depending on your testsuite name in the ``build.gradle.kts`` file of your project. For example::
+
+   val testOnDeviceMicroejVee by configurations.creating     
+
+- Declare the VEE in the project dependencies with your new configuration::
+
+   dependencies {
+       ...
+       testOnDeviceMicroejVee("com.mycompany:vee-port:1.0.0")
+   }
```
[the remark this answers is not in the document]
denisbedin: No, in the changes that I've done in the plugin, each test suite already has its own "xxxMicroejVeeClasspath" resolvable configuration, which by default depends on "testMicroejVee" if it is used or "microejVee". So the user has to create a "xxxMicroejVee" configuration only.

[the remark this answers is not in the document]
tdelhomenie-microej: Gradle recommends to not configure configuration with 2 roles (declarable + resolvable here), no?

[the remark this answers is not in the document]
denisbedin: Yes, I updated the doc to ensure that the user creates a declarable configuration.

tdelhomenie-microej (approved):

PR #1299, opened 2025-05-19
update path

rbabaousmail:
This MR updates the location where a linker script must be placed.


jerome-leroux:
Both path are good, aren't they? Do we need to update the documentation?

jerome-leroux (approved):

PR #1312, opened 2025-06-10
Update VEE port definition in glossary.rst

stephanemainchain:


frivieremicroej:
@stephanemainchain I'm not sure I understand the idea: is it to enforce the fact that there can be many VEE Ports for the same device ?

stephanemainchain:
it is to correct that the VEE port is not only device specific but board specific
- change the board = change the VEE port
- keep the device  but change the board => the VEE port must change

frivieremicroej (approved):

PR #1315, opened 2025-06-16
Update glossary.rst

stephanemainchain:
Fix typo.


frivieremicroej:
I remember the plural form was intentional at the time this glossary was introduced.
This enforces the fact that we feature a single container that manages multiple applications (no container duplication)
So I prefer to let it as-is.

stephanemainchain:
OK I agree.

frivieremicroej (approved):
align with MKT guidelines

PR #1309, opened 2025-06-04
improve documentation of the linear memory

frivieremicroej:

PR #1316, opened 2025-06-16
document memory limits in KF specification

frivieremicroej:


frivieremicroej:
NOTE this integration to main stream: (reviewed by #1309)

PR #1320, opened 2025-06-18
Clarify 'KF RAM Control: Managed Heap Configuration'

rlouedec-microej:
The working "overlap" is confusing (especially when reading from top to bottom).
Indeed, the diagrams clarify the meaning of "overlap", but, IMO, the important sentence is "This is an upper bound, not a reservation".

I suggest some rewording to clarify that.

Also, from this documentation, I understand that the features cannot allocate more than what would leave the kernel less than its reserved memory. It is correct?
Another implementation could use the same mechanism than described in the feature criticality section  (with Kernel having some criticality level) but I don't think this is the case, right?

### DIFF
--- a/KernelDeveloperGuide/kf.rst
+++ b/KernelDeveloperGuide/kf.rst
@@ -727,10 +727,10 @@ The following APIs allow to configure the Managed Heap usage:
   
 Especially, allocations from Features will fail (throw an OutOfMemoryError):
 
-    - when the feature-specific limit is reached, before the Managed heap is full;
-    - when the Managed heap is full, before the feature-specific limit is reached;
-    - when the Managed heap is almost full, so that the allocation would exceed the amount of memory that can be
-      allocated for Features (i.e. total Managed heap size minus memory reserved for Kernel).
+- when the feature-specific limit is reached, before the Managed heap is full;
+- when the Managed heap is full, before the feature-specific limit is reached;
+- when the Managed heap is almost full, so that the allocation would exceed the amount of memory that can be allocated
+  for Features (i.e. total Managed heap size minus memory reserved for Kernel).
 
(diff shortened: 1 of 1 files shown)


KernelDeveloperGuide/kf.rst
```diff
@@ -722,13 +722,15 @@ By default, the Kernel and Features can allocate without restriction, as long as
...
   
-  - Allocation may fail before the memory limit is reached.  
-  - Memory limits of multiple Features can overlap.  
-  - A Feature's memory limit can also overlap with the Kernel’s reserved memory.
+Especially, allocations from Features will fail (throw an OutOfMemoryError):
 
+- when the feature-specific limit is reached, before the Managed heap is full;
+- when the Managed heap is full, before the feature-specific limit is reached;
+- when the Managed heap is almost full, so that the allocation would exceed the amount of memory that can be allocated
+  for Features (i.e. total Managed heap size minus memory reserved for Kernel).
```
frivieremicroej, line 733: I don't like "Almost full", because it depends on the Kernel reserved memory size

```suggestion
- when the allocation would exceed the amount of memory that can be allocated
```


```suggestion
- when the allocation would exceed the amount of memory available to Features (i.e. Managed heap size minus memory reserved for the Kernel).
```

frivieremicroej:
> Another implementation could use the same mechanism than described in the feature criticality section (with Kernel having some criticality level) but I don't think this is the case, right?

the "criticality" notion is only for Features, not for the Kernel. The Kernel can always allocate objects unless the managed heap is full). It is obvious, but do you think it is worth mentioning that ?

rlouedec-microej:
@frivieremicroej No it's clear now, I just wanted to confirm this is the actual behavior (I don't think this is necessarily obvious w/o this doc).

frivieremicroej (approved):

PR #1341, opened 2025-07-09
M0269MD-612 Change the executable build flow section of the overview

gharelMicroEJ:


acolleux-microej:
The Executable Build Workflow section looks heavy. All the schematics are put one after another.

What about putting sub-section titles for every schematic?

Also, the Major Input or Output lengend looks weird. Especially when we begin by the "MicroEJ Resources to Build Applications and their Origins" that haven't introduced yet the build flow.
Could we rename it "Modules" or something like that?

After taking a closer look, we should remove the "MicroEJ" term when it's not necessary:
- MicroEJ Executable => Executable
- MicroEJ Example Applications => Example Applications
- MicroEJ VEE Port => VEE Port

PR #1346, opened 2025-07-15
Review link command description for Managed C

frivieremicroej:


ApplicationDeveloperGuide/managedc/compilation.rst
```diff
@@ -39,21 +39,29 @@ Linking a Wasm Module
 
 Once the C files are compiled into object files (``.o`` files),
 the next step is to link them together to produce a Wasm module (``.wasm`` file).
+Use ``clang`` as the linker driver to generate the final ``.wasm`` module. 
+It automatically calls ``wasm-ld``, the Wasm linker.
 
-For instance, if you have an object file named ``my_app.o`` in the ``src/main/resources`` folder,
-the following command line produces a Wasm module including the transitive dependencies to the :ref:`WASI Libc <wasm.wasi>`.
-All unresolved symbols will be expected to be defined by the Java host.
+The following options are recommended to ensure the produced ``.wasm`` binary is compatible with the MicroEJ VEE:
```
frivieremicroej, line 45:
```suggestion
The following options are recommended to ensure the produced ``.wasm`` binary is compatible with MICROEJ VEE:
```

ApplicationDeveloperGuide/managedc/compilation.rst
```diff
@@ -39,21 +39,29 @@ Linking a Wasm Module
...
+
+* ``-Wl,--allow-undefined``: Allows undefined symbols, which is necessary when the Wasm module references extern symbols defined in Java.
```
frivieremicroej, line 47:
```suggestion
* ``-Wl,--allow-undefined``: Allows undefined symbols, which is necessary when the Wasm module references extern symbols defined by the Java host.
```

ApplicationDeveloperGuide/managedc/compilation.rst
```diff
@@ -39,21 +39,29 @@ Linking a Wasm Module
...
-You have to specify the functions that can be called by the Java application using 
-the ``--export=[function_name]`` option for each function.     
+When your object file defines a ``main`` function, the linker automatically exports a function named ``_start``. 
+This function performs initialization tasks before invoking the user-defined ``main``. 
+You must bind the ``_start`` function — not ``main`` — to a Java method.
 
-When your object file declares a ``main`` function, the linker will automatically export a function named ``_start``. 
-This function will perform some initialization code before calling the user's main function. 
-You don't have to export the ``main`` function. You will just have to bind the function named ``_start`` to a Java method. 
+Additionally, any other functions that should be callable from the Java application must be explicitly exported using the ``-Wl,--export=[function_name]`` linker option (this does not apply to ``main``, which is invoked internally by ``_start``).
```
frivieremicroej, line 64:
```suggestion
Additionally, any other functions that should be callable from the Java Application must be explicitly exported using the ``-Wl,--export=[function_name]`` linker option (this does not apply to ``main``, which is invoked internally by ``_start``).  
```

ApplicationDeveloperGuide/managedc/compilation.rst
```diff
@@ -67,27 +75,24 @@ You don't have to export the ``main`` function. You will just have to bind the f
...
+For additional details and alternative configurations, see the :ref:`managedc.link.command_line_options` section.
 
 
 .. _managedc.link.command_line_options:
 
 Linker Options
 ``````````````
 
-Here are some useful linker options you might need:
+Here are other linker options you might need:
```
frivieremicroej, line 90: There are many references to this section that assume all declared options.
If we shrink the defined options, so I think we definitely should fully remove this section. I will suggest an other PR.

* `-Wl,--export=[function_name]` already declared above
* `-nostdlib` is already defined in <managedc.link.nostdlib> => we have to add `-Wl,--no-entry`
* `--initial-memory` is defined in the linear memory page

PR #1399, opened 2025-09-26
Update sandboxedAppSharedInterface.rst

gjadi-mej:
Remove occurrences of "remote" because the figure and text only mention "provider application", so keep this term


frivieremicroej (commented):
there are many other usages of remote in this page.
TDB

